PR #139, opened 2026-03-26
Fix: WebKit Inspector Protocol compatibility (navigate, target, async, cookies)

### LINKED ISSUE #137: WebKitClient: Fix navigation, target lifecycle, async evaluate, and cookie operations
## Problem

After fixing Target multiplexing (#128), several WebKit Inspector Protocol operations still fail due to protocol differences between Chrome DevTools Protocol and the actual WebKit Inspector Protocol exposed via ios-webkit-debug-proxy.

### Bug 1: `Page.navigate` not supported (CRITICAL)
```
ProtocolError: 'Page.navigate' was not found (code: -32601)
```
The WebKit Inspector Protocol does not expose `Page.navigate`. Navigation must use `Runtime.evaluate('window.location.href = url')` as a fallback.

### Bug 2: Target destroyed on navigation (CRITICAL)
When a page navigates (via click or JS), the current target is destroyed (`Target.targetDestroyed`) and a new target is created (`Target.targetCreated`). The client sets `activeTargetId = null` on destruction but does **not** auto-attach to the new target. All subsequent commands fail with "No active target".

### Bug 3: `Runtime.awaitPromise` returns empty object (HIGH)
```javascript
evaluate('new Promise(r => setTimeout(() => r(42), 100))') // returns {} instead of 42
```
The `Runtime.awaitPromise` response format may differ from what the client expects, or the method may not be supported in this protocol version.

### Bug 4: Cookie methods timeout (HIGH)
```
Page.setCookie timed out after 15000ms
Page.getCookies timed out after 15000ms
```
Cookie operations via `Page.setCookie`/`Page.getCookies` either don't exist in this protocol or use a different response format. Need to investigate alternative approaches (e.g., `Runtime.evaluate` with `document.cookie`).

### Fix 1: Navigate fallback
In `navigate()` method:
1. Try `Page.navigate` first
2. On -32601 error, fall back to `Runtime.evaluate('window.location.href = url')`
3. Wait for page load by polling `document.readyState` since `Page.loadEventFired` may also not work

```typescript
async navigate(options: NavigateOptions): Promise<NavigateResult> {
  const { url, waitUntil, timeout } = options;
  await this.enableDomain('Page').catch(() => {});
  
  try {
    await this.send('Page.navigate', { url });
  } catch (e) {
    if (e instanceof ProtocolError && e.code === -32601) {
      // Fallback: navigate via JS
      await this.evaluate(`window.location.href = ${JSON.stringify(url)}`);
    } else throw e;
  }
  
  // Wait for load via polling
  await this.waitForPageLoad(waitUntil, timeout);
  return { url, status: 200, loadTime: Date.now() - start };
}
```

### Fix 2: Auto-attach to new target after navigation
In `handleMessage()`, when `Target.targetCreated` fires:
- If `activeTargetId` is null (previous target was destroyed), auto-set the new page target
- Re-enable previously enabled domains on the new target

```typescript
if (msg.method === 'Target.targetCreated') {
  const info = msg.params?.targetInfo;
  if (info?.type === 'page') {
    if (!this.activeTargetId) {
      this.activeTargetId = info.targetId;
      // Re-enable domains
      const domains = [...this.enabledDomains];
      this.enabledDomains.clear();
      for (const domain of domains) {
        this.enableDomain(domain).catch(() => {});
      }
    }
    this.targetReadyResolve?.();
  }
  return;
}
```

### Fix 3: Async evaluate fallback
In `evaluate()`, when `Runtime.awaitPromise` fails or returns unexpected format:
1. Try `Runtime.awaitPromise` first
2. If it fails or returns empty, fall back to wrapping the expression in an async IIFE and polling

```typescript
// If awaitPromise returns empty/undefined, use polling fallback
if (awaited.result?.value === undefined && awaited.result?.type === 'object') {
  // Fallback: wrap in async and re-evaluate with returnByValue
  const wrappedResult = await this.send('Runtime.evaluate', {
    expression: `(async () => { return ${expression}; })().then(v => JSON.stringify(v))`,
    returnByValue: true,
    emulateUserGesture: true,
  });
  // Parse the stringified result
}
```

### Fix 4: Cookie operations via evaluate
Replace `Page.setCookie`/`Page.getCookies`/`Page.deleteCookie` with `Runtime.evaluate` based approaches:

```typescript
async getCookies(): Promise<Cookie[]> {
  try {
    return await this.send('Page.getCookies');
  } catch {
    // Fallback: parse document.cookie
    const raw = await this.evaluate<string>('document.cookie');
    return this.parseCookieString(raw);
  }
}

async setCookies(cookies: Cookie[]): Promise<void> {
  for (const cookie of cookies) {
    try {
      await this.send('Page.setCookie', cookie);
    } catch {
      await this.evaluate(`document.cookie = "${cookie.name}=${cookie.value}; path=${cookie.path || '/'}; domain=${cookie.domain || ''}"`);
    }
  }
}
```

## Files to Modify
- `src/webkit/client.ts` — All 4 fixes in one file

### Navigation (Bug 1)
- [ ] `navigate({url: 'https://example.com'})` completes without error
- [ ] `navigate` returns `{url, status, loadTime}` with reasonable values
- [ ] Page content is accessible via `evaluate('document.title')` after navigate
- [ ] `readPage()` returns content from the navigated page
- [ ] Multiple sequential navigations work (navigate → navigate → navigate)

### Target Lifecycle (Bug 2)
- [ ] After `click('a')` that triggers navigation, commands still work
- [ ] `activeTargetId` is automatically updated to new target
- [ ] Previously enabled domains are re-enabled on new target
- [ ] `evaluate()` works immediately after target transition
- [ ] `screenshot()` works after target transition
- [ ] Rapid navigation (click link → click another link) doesn't crash

### Async Evaluate (Bug 3)
- [ ] `evaluate('new Promise(r => setTimeout(() => r(42), 100))')` returns `42`
- [ ] `evaluate('Promise.resolve("hello")')` returns `"hello"`
- [ ] `evaluate('fetch("https://example.com").then(r => r.status)')` returns `200`
- [ ] Non-promise expressions still work: `evaluate('1+1')` returns `2`

### Cookies (Bug 4)
- [ ] `getCookies()` returns array (may be empty)
- [ ] `setCookies([{name:'test', value:'123', domain:'example.com', path:'/'}])` succeeds
- [ ] `getCookies()` includes the test cookie after set
- [ ] `clearCookies()` removes cookies
- [ ] `getCookies()` returns empty after clear

### Regression
- [ ] `npx tsc --noEmit` passes
- [ ] `npm run test` passes
- [ ] `npm run lint` passes (0 errors)
- [ ] Existing evaluate functionality unchanged (title, objects, errors)
- [ ] `screenshot()` still works
- [ ] `readPage()` still works

## Context
Discovered during Phase 1 verification (#122, sections 8-9, 11). These are all consequences of the WebKit Inspector Protocol being a subset of Chrome DevTools Protocol with different method availability.

shaun0927:
## Summary
- **Navigate**: Fall back to `window.location.href` when `Page.navigate` unavailable; poll `document.readyState` instead of `Page.loadEventFired`
- **Target lifecycle**: Auto-attach to new page target after navigation destroys the old one; re-enable domains on new target
- **Async evaluate**: Fallback when `Runtime.awaitPromise` returns empty objects
- **Cookies**: Fall back to `document.cookie` via evaluate when `Page.setCookie`/`Page.getCookies` unavailable

## Test plan
- [ ] `navigate('https://example.com')` succeeds
- [ ] `evaluate('document.title')` returns correct title after navigate
- [ ] Click-navigation doesn't break subsequent commands
- [ ] `evaluate('new Promise(r => setTimeout(() => r(42), 100))')` returns 42
- [ ] `getCookies()` returns array
- [ ] `setCookies([...])` succeeds
- [ ] `npx tsc --noEmit` passes
- [ ] `npm run test` passes
- [ ] `npm run lint` passes (0 errors)

Closes #137

### DIFF
--- a/src/webkit/client.ts
+++ b/src/webkit/client.ts
@@ -185,15 +185,22 @@ export class WebKitClient extends EventEmitter implements BrowserBackend {
     // Handle Target events (multiplexing protocol)
     if (msg.method === 'Target.targetCreated') {
       const info = msg.params?.targetInfo;
-      if (info?.type === 'page' && !this.activeTargetId) {
+      if (info?.type === 'page') {
+        // Always track the latest page target
+        // If old target still exists, it will be destroyed shortly
         this.activeTargetId = info.targetId;
         // Re-enable domains on new target (e.g., after navigation destroys old target)
         const domains = [...this.enabledDomains];
         this.enabledDomains.clear();
-        for (const domain of domains) {
-          this.enableDomain(domain).catch(() => {});
-        }
-        this.targetReadyResolve?.();
+        // Re-enable domains then signal target readiness
+        Promise.all(
+          domains.map(domain => this.enableDomain(domain).catch(err => {
+            console.error(`[WebKitClient] Failed to re-enable ${domain} on new target: ${(err as Error).message}`);
+          }))
+        ).then(() => {
+          this.targetReadyResolve?.();
+        });
+        return;  // Don't call targetReadyResolve here, wait for domains
       }
       return;
     }
@@ -381,18 +388,42 @@ export class WebKitClient extends EventEmitter implements BrowserBackend {
       await new Promise(r => setTimeout(r, 300));
       try {
         const readyState = await this.evaluate<string>('document.readyState');
-        if (waitUntil === 'domcontentloaded' && (readyState === 'interactive' || readyState === 'complete')) break;
-        if (waitUntil === 'load' && readyState === 'complete') break;
-        if (!waitUntil && readyState === 'complete') break;
+        if (waitUntil === 'networkidle') {
+          // networkidle not directly detectable via polling
+          // Fall back to 'complete' readyState + extra delay
+          if (readyState === 'complete') {
+            await new Promise(r => setTimeout(r, 500)); // extra settle time
+            break;
+          }
+        } else if (waitUntil === 'domcontentloaded') {
+          if (readyState === 'interactive' || readyState === 'complete') break;
+        } else if (waitUntil === 'load') {
+          if (readyState === 'complete') break;
+        } else {
+          if (readyState === 'complete') break;
+        }
       } catch {
         // Target may be transitioning during navigation, keep polling
         continue;
       }
     }
 
+    // P0-1: Check if we actually broke out or timed out
+    const finalReadyState = await this.evaluate<string>('document.readyState').catch(() => '');
+    const expectedState = waitUntil === 'domcontentloaded' ? 'interactive' : 'complete';
+    if (finalReadyState !== 'complete' && finalReadyState !== expectedState) {
+      throw new TimeoutError(`Navigation timeout after ${navTimeout}ms (readyState: ${finalReadyState})`);
+    }
+
+    // P0-2: Try to get real HTTP status from Performance API
+    const currentUrl = await this.evaluate<string>('document.URL').catch(() => options.url);
+    const status = await this.evaluate<number>(
+      `(function() { try { var e = performance.getEntriesByType('navigation')[0]; return e ? e.responseStatus || 200 : 200; } catch(ex) { return 200; } })()`
+    ).catch(() => 200);
+
     return {
-      url: options.url,
-      status: 200,
+      url: currentUrl,
+      status,
       loadTime: Date.now() - startTime,
     };
   }
@@ -465,6 +496,9 @@ export class WebKitClient extends EventEmitter implements BrowserBackend {
       }
 
       // Fallback: re-evaluate with async wrapper that JSON-stringifies the result
+      // NOTE: This fallback re-evaluates the expression. Expressions with side effects
+      // (network requests, mutations) will execute twice. This is a known limitation
+      // of the WebKit Inspector Protocol's incomplete Runtime.awaitPromise support.
       const fallback = await this.send<{
         result: { type: string; subtype?: string; value?: unknown; objectId?: string };
         wasThrown: boolean;
@@ -558,6 +592,8 @@ export class WebKitClient extends EventEmitter implements BrowserBackend {
       return cookies;
     } catch {
       // Fallback: parse document.cookie via evaluate
+      // Fallback: document.cookie cannot access httpOnly cookies.
+      // Results will be incomplete for httpOnly cookies.
       const raw = await this.evaluate<string>('document.cookie');
       if (!raw) return [];
       return raw.split(';').map(pair => {


shaun0927 (commented):
## PR #139: Fix: WebKit Inspector Protocol compatibility

- [ ] **[P0]** Navigate polling loop silently succeeds on timeout — when the while loop expires, execution falls through and returns `{status: 200}` as if navigation succeeded. Must throw `TimeoutError` after the loop. — `src/webkit/client.ts:380-397` (Confidence: 98/100)

- [ ] **[P0]** Navigate returns hardcoded `status: 200` — should attempt to read actual status via `performance.getEntriesByType('navigation')[0]?.responseStatus` or document the limitation. — `src/webkit/client.ts:394` (Confidence: 92/100)

- [ ] **[P0]** `Target.targetCreated` handler swallows domain re-enable failures via `.catch(() => {})` — if re-enabling fails, subsequent commands fail with cryptic errors. No retry, no error propagation, `targetReadyResolve` called regardless. — `src/webkit/client.ts:194` (Confidence: 90/100)

### P1 — Must Fix

- [ ] **[P1]** `Target.targetCreated` only fires when `!activeTargetId` — race condition if new target appears before old one destroyed. — `src/webkit/client.ts:188` (Confidence: 82/100)
- [ ] **[P1]** Async evaluate fallback re-executes the expression — side effects run twice. Should use objectId from first evaluation. — `src/webkit/client.ts:472` (Confidence: 95/100)
- [ ] **[P1]** JSON.stringify/parse roundtrip loses undefined, BigInt, Date, Map/Set, Infinity/NaN — document limitation. — `src/webkit/client.ts:472,496` (Confidence: 88/100)
- [ ] **[P1]** `waitUntil: 'networkidle'` has no branch in polling loop — spins for full timeout then silently succeeds. — `src/webkit/client.ts:384-386` (Confidence: 97/100)
- [ ] **[P1]** Cookie fallback via `document.cookie` cannot read httpOnly cookies — silently incomplete results. — `src/webkit/client.ts:559-574` (Confidence: 95/100)

### P2 — Improve

- [ ] **[P2]** 300ms polling interval hardcoded — extract to constant
- [ ] **[P2]** `clearCookies` tries `Page.deleteAllCookies` which doesn't exist — always falls through
- [ ] **[P2]** evaluate fallback returns `undefined as T` silently
- [ ] **[P2]** getCookies fallback omits sameSite field
- [ ] **[P2]** clearCookies JS fallback doesn't clear subdomain/non-root-path cookies

### Summary
| Priority | Count |
|----------|-------|
| P0 | 3 |
| P1 | 5 |
| P2 | 5 |

### Verdict
**REQUEST_CHANGES** — Fix P0s (timeout handling, hardcoded status, domain re-enable) and P1s before merge.

Issue #129, opened 2026-03-26
Auto-detect webinspectord_sim.socket path for Xcode 26+


shaun0927:
## Problem

`ios-webkit-debug-proxy` cannot find the iOS Simulator's Web Inspector socket at its default location. The proxy starts but lists zero devices, making Safari connection impossible without manual `-s` flag.

### Expected behavior
`ios_webkit_debug_proxy` automatically discovers the simulator and lists Safari targets.

### Actual behavior
```
$ ios_webkit_debug_proxy &
$ curl http://localhost:9221
iOS Devices: (empty list)
```

## Root Cause

The Web Inspector daemon socket path changed in Xcode 26.4:

| Version | Socket Path |
|---------|-------------|
| Xcode ≤ 15 | `/private/tmp/com.apple.launchd.*/com.apple.webinspectord_sim.socket` |
| **Xcode 26.4** | **`/private/var/tmp/com.apple.launchd.*/com.apple.webinspectord_sim.socket`** |

The `ios-webkit-debug-proxy` default path (`localhost:27753` or the `/private/tmp/` glob) no longer matches.

### Workaround
Manually specify the socket path:
```bash
ios_webkit_debug_proxy -s unix:/private/var/tmp/com.apple.launchd.<hash>/com.apple.webinspectord_sim.socket
```

### 1. Add socket auto-detection in `src/simulator/manager.ts` or new `src/simulator/proxy.ts`

```typescript
async function findWebInspectorSocket(): Promise<string | null> {
  const searchPaths = [
    '/private/var/tmp',  // Xcode 26+
    '/private/tmp',      // Xcode ≤ 15
  ];

  for (const base of searchPaths) {
    const dirs = await fs.readdir(base).catch(() => []);
    for (const dir of dirs) {
      if (!dir.startsWith('com.apple.launchd.')) continue;
      const socketPath = path.join(base, dir, 'com.apple.webinspectord_sim.socket');
      try {
        await fs.access(socketPath);
        return `unix:${socketPath}`;
      } catch { continue; }
    }
  }
  return null;
}
```

### 2. Auto-start proxy with correct socket in `device_boot` tool

After booting a simulator, automatically:
1. Detect the web inspector socket path
2. Start `ios_webkit_debug_proxy` with `-s <socket>` flag
3. Verify connection with `curl http://localhost:9222/json`

### 3. Update `doctor` command

Add a check for:
- `ios_webkit_debug_proxy` binary availability
- Web inspector socket existence
[...]

PR #141, opened 2026-03-27
fix: proxy port conflict detection, env var support, and documentation

shaun0927:
## Summary
- Fix port conflict detection bug where a second `WebInspectorProxy` instance could start without error on an already-used port — now uses `net.connect()` (connect-test) instead of `net.createServer()` (bind-test) and also checks for running `ios_webkit_debug_proxy` processes via `pgrep`
- Add `OPENSAFARI_PROXY_PORT` environment variable support (resolution order: explicit option > env var > default 9322)
- Add JSDoc documenting default port (9322, chosen to avoid conflict with openchrome on 9222)

Closes partially #129 (Port Conflict Avoidance section)

## Test plan
- [ ] Start proxy, then attempt second instance on same port — should throw with helpful error message
- [ ] Set `OPENSAFARI_PROXY_PORT=9400` and verify proxy starts on port 9400
- [ ] `npm run test` passes

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/simulator/proxy.ts
+++ b/src/simulator/proxy.ts
@@ -153,9 +153,14 @@ export class WebInspectorProxy {
   private isPortInUse(port: number): Promise<boolean> {
     return new Promise(resolve => {
       const socket = net.connect({ port, host: '127.0.0.1' });
+      socket.setTimeout(2000);
       socket.once('connect', () => {
         socket.destroy();
         resolve(true);
+      });
+      socket.once('timeout', () => {
+        socket.destroy();
+        resolve(false);
       });
       socket.once('error', () => {
         socket.destroy();


shaun0927 (commented):
## PR #141: fix: proxy port conflict detection, env var support, and documentation

None found.

### P1 — Must Fix (should fix in this PR)
- [ ] **[P1]** `isPortInUse` socket has no timeout — `net.connect()` can hang indefinitely if the target accepts but never responds — `src/simulator/proxy.ts:153-160` (Confidence: 75/100)
  - Fix: Add `setTimeout` on the socket (e.g. 2000ms) and resolve `true` on timeout, or use the `timeout` option of `net.connect`.

### P2 — Improve (can be follow-up)
- [ ] **[P2]** `pgrep -x` matches exact process name only — on some macOS setups the binary name in the process table may differ (e.g. truncated). Consider `pgrep -f ios_webkit_debug_proxy` for broader matching. — `src/simulator/proxy.ts:170`
  - Suggestion: Use `-f` flag for pattern match or verify `-x` works on all target macOS versions.

### Summary
| Priority | Count |
|----------|-------|
| P0 | 0 |
| P1 | 1 |
| P2 | 1 |

### Verdict
**REQUEST_CHANGES** — P0 = 0, P1 = 1. One fixable issue (socket timeout).

### Merge Notes
- **Conflicts with PR #142** — both rewrite `proxy.ts`. These cannot be merged together. Recommend merging #141 first (conservative approach that preserves existing architecture) and rebasing #142 afterward.
- No conflicts with #143 or #145.

PR #142, opened 2026-03-27
fix: auto-start proxy on device_boot with PID lifecycle tracking

shaun0927:
## Summary
- Integrate `WebInspectorProxy` auto-start into `device_boot` tool — proxy launches automatically after simulator boot with detected socket path
- Add module-level singleton `getSharedProxy()` so the proxy instance is shared across tool calls
- Track proxy PID and register `process.on('exit')` cleanup handler for graceful shutdown
- Proxy failure is non-blocking — boot succeeds even if proxy fails to start
- Boot response now includes `proxy: { running, pid }` status

Closes partially #129 (Integration with device_boot section)

## Test plan
- [ ] `device_boot` auto-starts proxy after simulator boot
- [ ] Response JSON includes proxy status with PID
- [ ] Proxy is cleaned up on process exit
- [ ] Boot succeeds even if proxy start fails (e.g., no `ios_webkit_debug_proxy` installed)
- [ ] `npm run test` passes

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/simulator/index.ts
+++ b/src/simulator/index.ts
@@ -8,5 +8,5 @@ export { SimulatorPool, InsufficientResourcesError } from './pool';
 export type { PooledSimulator, SimulatorPoolOptions } from './pool';
 export { BatchExecutor } from './batch';
 export type { BatchResult } from './batch';
-export { WebInspectorProxy } from './proxy';
+export { WebInspectorProxy, getSharedProxy } from './proxy';
 export type { ProxyOptions } from './proxy';


shaun0927 (commented):
## Code Review — PR #142

**Verdict: APPROVE** ✅

All 7 previously reported issues (2× P0, 3× P1, 2× P2) have been fixed in commit eb7c13a.

### Verified fixes:
- **P0**: Default port restored to 9322 (was 9221), dynamic `-c` flag with device port range
- **P1**: Safety checks restored (which, port-in-use, pgrep), env var `OPENSAFARI_PROXY_PORT` support, single error listener
- **P2**: Async `stop()` with SIGKILL fallback, `getSharedProxy` exported from barrel

### Remaining (non-blocking):
- [MEDIUM] `parseInt` on env var has no NaN guard — edge case only
- [LOW] No unit tests for proxy module

### Anti-pattern scan: CLEAN
Build: PASS | Tests: PASS | Lint: PASS

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

PR #155, opened 2026-03-27
fix: derive device-list port from OPENSAFARI_PROXY_PORT (#147)

### LINKED ISSUE #147: Multi-session proxy conflict: pgrep system-wide check blocks parallel Claude Code sessions
## Problem

When multiple Claude Code sessions use opensafari simultaneously, the second session fails to start the WebInspector proxy because `proxy.ts` uses a **system-wide `pgrep -x`** check that rejects ANY running `ios_webkit_debug_proxy` process — even when the new session requests a different port.

This makes opensafari **single-session only** in practice, despite having multi-device infrastructure (`SimulatorPool`, `SessionManager`) already implemented.

### Reproduction

```bash
# Terminal A: Claude Code session boots a device (proxy starts on port 9322)
> device_boot iPhone 17e
# => proxy: { running: true, pid: 12345 }

# Terminal B: Another Claude Code session tries to boot
> device_boot iPhone 17e
# => proxy: { running: false, pid: null }
# stderr: [device_boot] Failed to start WebInspector proxy:
#   Error: Another ios_webkit_debug_proxy process is already running.

# Even with a different port:
OPENSAFARI_PROXY_PORT=9500 node -e "new WebInspectorProxy({port:9500}).start()"
# => Same error — pgrep blocks before port check runs
```

### Expected behavior

Multiple Claude Code sessions should be able to share or independently manage `ios_webkit_debug_proxy` processes without conflicting.

### Actual behavior

The second session's `device_boot` silently fails to start the proxy (`proxy: {running: false, pid: null}`), leaving all subsequent tools (`navigate`, `click`, `screenshot`, etc.) broken with "Safari not connected".

## Root Cause Analysis

The conflict has **three layered barriers**:

### Layer 1 (P0): System-wide `pgrep` in `isProxyAlreadyRunning()`

```typescript
// proxy.ts:205-211
private async isProxyAlreadyRunning(): Promise<boolean> {
  try {
    await execFileAsync('pgrep', ['-x', 'ios_webkit_debug_proxy']);
    return true;   // ANY process with this name → block
  } catch {
    return false;
  }
}
```

This fires at `start()` line 68, **before** the port-specific `isPortInUse()` check at line 77. Even with `OPENSAFARI_PROXY_PORT=9500`, the method throws because it detects another session's proxy on a completely different port.

### Layer 2 (P1): `process.on('exit')` kills shared proxy

```typescript
// proxy.ts:241-247
process.on('exit', () => {
  if (_sharedProxy?.running) {
    try { _sharedProxy['process']?.kill('SIGTERM'); } catch {}
  }
});
```

When Session A's MCP process exits, it kills the proxy — taking down Session B's connectivity too. There is no ownership check.

### Layer 3 (P1): No proxy cleanup in `device_shutdown`

`device-shutdown.ts` has zero references to proxy. Shutdown only stops the simulator; the proxy process is orphaned until the MCP server process itself exits.

### Additional Issues Found

| Severity | Issue | Location |
|----------|-------|----------|
| P1 | Doctor `PROXY_PORTS = [9222, 9322]` doesn't include device-list ports (9221, 9321) — `proxyReachable` always false | `xcode-check.ts:11` |
| P2 | Default port 9322 and `OPENSAFARI_PROXY_PORT` not documented in README/docs | Only in JSDoc at `proxy.ts:11-16` |
| P2 | Port conflict error at `proxy.ts:78-84` is unreachable when pgrep check fires first | `proxy.ts:68` vs `proxy.ts:77` order |

### Phase 1: Reuse existing healthy proxy (P0 fix)

Replace the exclusive `pgrep` check with a **detect-and-reuse** pattern:

```typescript
async start(): Promise<void> {
  if (this._running) return;

  // 1. Check if OUR target port is already in use
  const portInUse = await this.isPortInUse(this._port);
  if (portInUse) {
    // Check if it's a healthy ios_webkit_debug_proxy we can reuse
    const isHealthy = await this.isProxyHealthy();
    if (isHealthy) {
      this._running = true;
      this._reusing = true;  // Don't kill on exit
      return;
    }
    throw new Error(
      `Port ${this._port} already in use by another process. ` +
      `Use a different port via OPENSAFARI_PROXY_PORT env var ` +
      `(e.g. OPENSAFARI_PROXY_PORT=${this._port + 100}).`
    );
  }

  // 2. Port is free — start a new proxy
  // ... existing spawn logic ...
}
```

Key changes:
- **Remove** `isProxyAlreadyRunning()` (system-wide `pgrep`)
- **Add** `isProxyHealthy()` — HTTP GET to device-list port, check for "iOS Devices" response
- **Add** `_reusing` flag — if reusing another session's proxy, do NOT kill on exit
- **Update** exit handler to only kill self-owned proxies

### Phase 2: Proxy lifecycle integration

- `device_shutdown` should call `getSharedProxy().stop()` when the last device is shut down
- Track proxy ownership: only the process that spawned it may kill it (use PID file or flag)

### Phase 3: Fix Doctor connectivity check

```typescript
// xcode-check.ts — fix port list
const PROXY_DEVICE_LIST_PORTS = [9221, 9321]; // actual device-list ports
const PROXY_DEVICE_PORTS = [9222, 9322];      // device connection ports
```

Check device-list ports for "iOS Devices" response, and device ports for JSON target list.

### Phase 4: Documentation

Add to README:
- Default proxy port: 9322 (device), 9321 (device list)
- `OPENSAFARI_PROXY_PORT` env var for customization
- Multi-session usage notes

### Q1: Reuse vs. per-session isolation?

| Approach | Pros | Cons |
|----------|------|------|
| **Reuse existing proxy** (Phase 1) | Minimal code change, unblocks most use cases | Session B depends on Session A's proxy staying alive |
| **Per-session port isolation** | True independence | Multiple proxy instances, port coordination, more RAM |
| **Shared daemon with lockfile** | Clean ownership | Stale lockfiles on crash, complexity |

**Recommendation**: Phase 1 (reuse) covers 90% of use cases. Per-session isolation can follow as Phase 5 if needed.

### Q2: Should `SessionManager` be wired into tools?

`SessionManager` (`session-manager.ts`) already has multi-device connection tracking (`connections: Map<string, BrowserBackend>`) but zero tools use it — they all call the global `getWebKitClient()`. Wiring `SessionManager` into tools would enable true multi-device support within a single process, but requires changing every tool's client acquisition pattern. This is a larger architectural change and may be better tracked separately.

### Q3: Should the exit handler be removed entirely?

The `process.on('exit')` handler at `proxy.ts:242-247` is a safety net but dangerous in multi-session. Options:
- A) Keep but add ownership check (`if (this._ownedByThisProcess)`)
- B) Remove entirely, rely on `device_shutdown` for cleanup
- C) Replace with a PID file (`/tmp/opensafari-proxy.pid`) for cross-process coordination

**Recommendation**: Option A — simplest and safest.

## Verification Checklist

> All items must pass on `develop` branch (Xcode 26.4, iOS 26.4) before merge.

#### Single-session baseline (regression)
- [x] `device_boot` starts proxy when no proxy is running — returns `proxy: {running: true, pid: <number>}`
- [x] After boot, `navigate` / `read_page` / `screenshot` tools work correctly
- [x] `device_shutdown` gracefully stops the simulator
- [x] After MCP process exits, the self-owned proxy is terminated (no orphan)

#### Multi-session coexistence
- [x] Session A boots device and starts proxy on default port 9322
- [x] Session B boots device — detects existing proxy, reuses it (`_reusing = true`)
- [x] Session B's `navigate` / `screenshot` tools work while Session A's proxy is alive
- [ ] Session A exits — proxy is NOT killed (Session B still using it, or ownership transferred)
- [ ] Session B continues to function after Session A exits
- [ ] Session B exits — proxy is killed (last user)

#### Different-port scenario
- [ ] `OPENSAFARI_PROXY_PORT=9500` starts proxy on port 9500 (no conflict with existing 9322)
- [ ] Two proxies on different ports can coexist simultaneously
- [ ] Each session's tools connect to its own proxy's port

#### Port conflict handling
- [x] When port is occupied by a non-proxy process, error message suggests `OPENSAFARI_PROXY_PORT`
- [x] When port is occupied by a healthy proxy, proxy is reused silently
- [x] When port is occupied by a dead/unresponsive proxy, error is thrown with clear instructions

#### pgrep removal validation
- [x] `isProxyAlreadyRunning()` is removed or replaced
- [x] No remaining system-wide `pgrep` calls in proxy.ts
- [x] Proxy detection is now port-specific, not process-name-based

#### device_shutdown cleanup
- [x] `device_shutdown` calls `getSharedProxy().stop()` when proxy was self-started
- [x] `device_shutdown` does NOT stop proxy if it was reused from another session
- [x] Orphan proxy processes do not accumulate after repeated boot/shutdown cycles
- [x] Running `device_shutdown` → `device_boot` cycle 3 times produces zero orphans

#### Ownership tracking
- [x] Process that started the proxy is recorded (PID file or in-memory flag)
- [x] `process.on('exit')` only kills proxy if current process is the owner
- [x] Non-owner process exit does not affect proxy

### Phase 3: Doctor connectivity fix

- [x] `checkProxyReachable()` checks device-list ports (9221, 9321) for "iOS Devices" response
- [ ] `checkProxyReachable()` checks device ports (9222, 9322) for JSON target list
- [x] `proxyReachable: true` when opensafari proxy is running on default port 9321/9322
- [x] `proxyReachable: true` when traditional proxy is running on 9221/9222
- [x] `proxyReachable: false` when no proxy is running (no false positive)
- [x] Doctor output includes which port the proxy was found on

### Phase 4: Documentation

- [x] README documents default ports: 9322 (device), 9321 (device list)
- [x] README documents `OPENSAFARI_PROXY_PORT` env var with example
- [x] README documents multi-session usage (shared proxy behavior)
- [x] CLAUDE.md updated if proxy-related instructions change

### Build & CI
- [x] `npm run build` succeeds with zero errors
- [x] `npm run test` passes (all existing tests)
- [x] `npm run lint` passes with zero errors
- [x] No new TypeScript `any` types introduced
- [ ] New proxy reuse logic has unit tests (mock pgrep removal, port detection)

### Integration test (manual)
- [ ] Open two Claude Code terminals side-by-side
- [ ] Terminal A: `device_boot iPhone 17e` → success with proxy
- [ ] Terminal B: `device_boot iPhone 17e` → success (reuses proxy)
- [ ] Terminal A: `navigate` to `https://example.com` → works
- [ ] Terminal B: `navigate` to `https://wikipedia.org` → works
- [ ] Terminal A: close Claude Code session
- [ ] Terminal B: `screenshot` → still works (proxy alive)
- [ ] Terminal B: `device_shutdown` → proxy cleaned up
- [ ] Verify zero orphan `ios_webkit_debug_proxy` processes: `pgrep -ax ios_webkit_debug_proxy` returns empty

## References

- #129 — Original socket detection issue (contains runtime verification notes)
- `src/simulator/proxy.ts` — Proxy lifecycle (`pgrep` at L205, exit handler at L242)
- `src/tools/device-boot.ts` — Boot + proxy integration (L23-41)
- `src/tools/device-shutdown.ts` — No proxy reference
- `src/simulator/xcode-check.ts` — Doctor checks (port list at L11)
- `src/session-manager.ts` — Unused multi-device session manager
- `src/mcp-server.ts:32-40` — Global `webkitClient` singleton

shaun0927:
## Problem

When `OPENSAFARI_PROXY_PORT=9500` is set, `_port` becomes 9500 but `_deviceListPort` remained hardcoded at 9321. This caused two bugs:

1. **False reuse detection**: Session B with port 9500 would detect Session A's proxy on 9321 as its own (same `_deviceListPort`), skip starting a new proxy, then fail because the actual device port didn't match.
2. **Port mismatch**: `WebKitClient({ port: 9500 })` fails because the proxy was serving devices on 9322, not 9500.
3. **Port conflict**: Two proxy instances with different device ports could not coexist since they would both try to claim port 9321 as their device-list port.

Fixes #147.

## Solution

Derive `_deviceListPort` as `port - 1` when no explicit override is provided. Resolution order (highest to lowest precedence):

1. Explicit `deviceListPort` constructor option
2. `OPENSAFARI_PROXY_DEVICE_LIST_PORT` environment variable (new)
3. `port - 1` (derived from resolved `_port`)

This preserves backwards compatibility — the defaults (port=9322, deviceListPort=9321) are unchanged — while allowing multi-port coexistence:

| `OPENSAFARI_PROXY_PORT` | Resolved `port` | Resolved `deviceListPort` |
|---|---|---|
| _(unset)_ | 9322 | 9321 |
| `9500` | 9500 | 9499 |
| `9500` + `OPENSAFARI_PROXY_DEVICE_LIST_PORT=9400` | 9500 | 9400 |

## Changes

- `src/simulator/proxy.ts`: Updated constructor to derive `_deviceListPort` from `_port`; updated `ProxyOptions` JSDoc and class-level JSDoc.
- `tests/unit/proxy-port.test.ts`: Added 5 unit tests covering all resolution cases.

## Test plan

- [ ] `npm run build` — compiles without errors
- [ ] `npm test` — all 7 tests pass (5 new + 2 existing)
- [ ] Default behaviour unchanged: `new WebInspectorProxy()` → port 9322, deviceListPort 9321
- [ ] Custom port: `new WebInspectorProxy({ port: 9500 })` → deviceListPort 9499
- [ ] `OPENSAFARI_PROXY_PORT=9500` → deviceListPort 9499
- [ ] `OPENSAFARI_PROXY_DEVICE_LIST_PORT=9400` overrides derived value
- [ ] Explicit `deviceListPort` option wins over all env vars

### DIFF
--- a/src/simulator/proxy.ts
+++ b/src/simulator/proxy.ts
@@ -53,11 +53,15 @@ export class WebInspectorProxy {
   private _reusing = false;
 
   constructor(private options: ProxyOptions = {}) {
-    const envPort = process.env.OPENSAFARI_PROXY_PORT
+    const rawPort = process.env.OPENSAFARI_PROXY_PORT
       ? parseInt(process.env.OPENSAFARI_PROXY_PORT, 10)
       : undefined;
-    const envDeviceListPort = process.env.OPENSAFARI_PROXY_DEVICE_LIST_PORT
+    const envPort = rawPort !== undefined && !Number.isNaN(rawPort) ? rawPort : undefined;
+    const rawDeviceListPort = process.env.OPENSAFARI_PROXY_DEVICE_LIST_PORT
       ? parseInt(process.env.OPENSAFARI_PROXY_DEVICE_LIST_PORT, 10)
+      : undefined;
+    const envDeviceListPort = rawDeviceListPort !== undefined && !Number.isNaN(rawDeviceListPort)
+      ? rawDeviceListPort
       : undefined;
     this._port = options.port ?? envPort ?? 9322;
     this._deviceListPort = options.deviceListPort ?? envDeviceListPort ?? (this._port - 1);
--- a/tests/unit/proxy-port.test.ts
+++ b/tests/unit/proxy-port.test.ts
@@ -58,4 +58,19 @@ describe('WebInspectorProxy port derivation', () => {
     expect(proxy.port).toBe(9500);
     expect(proxy.deviceListPort).toBe(9200);
   });
+
+  it('non-numeric OPENSAFARI_PROXY_PORT falls back to default 9322', () => {
+    process.env.OPENSAFARI_PROXY_PORT = 'abc';
+    const proxy = new WebInspectorProxy();
+    expect(proxy.port).toBe(9322);
+    expect(proxy.deviceListPort).toBe(9321);
+  });
+
+  it('non-numeric OPENSAFARI_PROXY_DEVICE_LIST_PORT falls back to derived port - 1', () => {
+    process.env.OPENSAFARI_PROXY_PORT = '9600';
+    process.env.OPENSAFARI_PROXY_DEVICE_LIST_PORT = 'abc';
+    const proxy = new WebInspectorProxy();
+    expect(proxy.port).toBe(9600);
+    expect(proxy.deviceListPort).toBe(9599);
+  });
 });


shaun0927 (commented):
## Code Review — PR #155

**P0: 0 | P1: 0 (2 fixed) | P2: 1**

### P1 fixes applied (commit 727fdfd):
1. **NaN guard on `parseInt`**: Added `Number.isNaN` check for both `OPENSAFARI_PROXY_PORT` and `OPENSAFARI_PROXY_DEVICE_LIST_PORT` env vars to prevent silent `NaN` propagation
2. **Test coverage**: Added 2 tests for non-numeric env var fallback behavior (total: 9 pass)

### Remaining P2 (non-blocking):
- Port boundary edge case: `port = 1` yields `deviceListPort = 0` (unrealistic in practice)

**Verdict: APPROVE** — Core logic correct, backward-compatible, well-tested after fixes.

Issue #260, opened 2026-03-30
E2E: Validate console_log monitoring and error event capture


shaun0927:
## Problem

The `console_log` tool subscribes to console and error events from Safari, but has never been tested on real pages generating actual console output:

- `console.log()`, `console.warn()`, `console.error()` messages from page JavaScript
- Unhandled promise rejections
- Network errors (failed fetch, 404 resources)
- JavaScript runtime errors (TypeError, ReferenceError)
- Console output from third-party scripts (analytics, ads)
- Large volume console output (performance and buffer handling)

This tool is part of #200 (console/network/error events as MCP tools), but basic console monitoring already exists in the codebase and needs e2e validation.

## Proposed Solution

1. **Basic capture**: Navigate to page with `console.log('test')`, verify message captured
2. **Error levels**: Generate log/warn/error, verify severity correctly categorized
3. **JS errors**: Navigate to page with intentional TypeError, verify error captured with stack trace
4. **Network errors**: Load page with broken image/script, verify network error event captured
5. **Volume test**: Generate 100 console messages rapidly, verify all captured without loss
6. **Filtering**: Verify console output doesn't interfere with MCP JSON-RPC on stdout

## Priority

**P1** — Console monitoring is essential for debugging. Missed errors mean invisible failures in QA workflows.

## Acceptance Criteria

- [x] All console levels (log, warn, error) captured from real pages
- [x] JavaScript runtime errors captured with stack traces
- [x] Network errors surfaced as events
- [x] High-volume output doesn't cause memory issues or message loss
- [x] Console output routed to stderr (not stdout) to preserve MCP protocol
- [x] Events include page URL and timestamp context

## E2E Verification Results (2026-03-31)

All 6 criteria verified on real Safari in iOS 26.4 Simulator (iPhone 17 Pro):

| Criterion | Result | Detail |
|-----------|--------|--------|
| Console levels | ✅ PASS | All 3 levels (log, warn, error) captured from real pages |
| JS errors with stack | ✅ PASS | TypeError + ReferenceError captured via console error level |
| Network errors | ✅ PASS | 3 events: 2 requests + 1 response (404 error) |
| High-volume | ✅ PASS | 100/100 messages captured without loss |
| Stderr routing | ✅ PASS | No console.log() in tool/server code; stdout is JSON-RPC only |
| URL & timestamp | ✅ PASS | Timestamps present + URL context in network entries |

PR #302, opened 2026-03-31
fix: error_log captures Safari JS errors via Console domain fallback

shaun0927:
## Summary
- Add Console domain fallback in `error_log` tool to capture JS runtime errors on Safari
- Safari/WebKit routes TypeError, ReferenceError, etc. through `Console.messageAdded` (error level) instead of `Runtime.exceptionThrown`
- Message-based deduplication prevents double-counting when both channels fire
- Only captures messages matching JS error patterns (not general `console.error()` calls)

## Context
E2E verification of #260 showed `error_log` captured 0 errors on real Safari while `console_log` captured them as error-level messages. This is the tool-level fix complementing the WebKitClient fix in #301.

## Test plan
- [x] `npm run build` passes
- [x] All 449 tests pass (36 suites)
- [ ] E2E: `error_log` captures TypeError/ReferenceError on Safari
- [ ] E2E: No duplicate entries when both Runtime and Console channels fire

### DIFF
--- a/src/tools/error-log.ts
+++ b/src/tools/error-log.ts
@@ -40,13 +40,16 @@ export function registerErrorLogTool(server: MCPServer): void {
 
       if (action === 'start') {
         if (attachedClient !== client) {
-          const recentMessages = new Set<string>();
+          const DEDUP_WINDOW_MS = 500;
+          const recentMessages = new Map<string, number>();
           const pushError = (entry: ErrorEntry) => {
-            const key = entry.message;
-            if (recentMessages.has(key)) return;
-            recentMessages.add(key);
+            const key = `${entry.message}|${entry.source ?? ''}|${entry.line ?? ''}`;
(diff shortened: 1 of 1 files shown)


shaun0927 (commented):
Code review: P0=0, P1=2, P2=2

**P1:** Dedup Set permanently suppresses repeated errors — a `TypeError` that fires 500 times is recorded only once. For a monitoring tool, this is data loss. Fix: use time-windowed dedup (Map<string, number> with 500ms window) to prevent cross-channel double-fire while preserving repeated error visibility.

**P1:** Both `console_log` and `error_log` subscribe to Console.messageAdded, so errors appear in both tools. Document the overlap or coordinate.

P2s: `onConsole` guard unnecessary (required method); `(client as any)` casts bypass type safety.

Will fix the dedup P1 before merging.

Issue #426, opened 2026-04-13
feat: Flutter-specific QA detector suite


shaun0927:
## Problem

opensafari has 13 web-focused QA detectors (`qa_touch_targets`, `qa_safe_area`, `qa_horizontal_overflow`, etc.) that run JavaScript in Safari. None of these work with Flutter apps because there is no DOM. Flutter apps need equivalent quality checks using the accessibility tree and screenshot analysis.

### 1. `qa_flutter_touch_targets` (High Priority)

**What**: Verify all interactive elements meet minimum 48x48dp tap target size (Material Design / WCAG 2.5.8).

**How**: 
```typescript
// Read accessibility tree → filter interactive elements (buttons, links, inputs)
// → check frame.width >= 48 && frame.height >= 48
const tree = await accessibilityBridge.dumpTree({ deviceId });
const interactiveRoles = ['AXButton', 'AXLink', 'AXTextField', 'AXCheckBox', 'AXSwitch'];
const violations = findNodes(tree, node => 
  interactiveRoles.includes(node.role) && 
  (node.frame.width < 48 || node.frame.height < 48)
);
```

### 2. `qa_flutter_semantics_coverage` (High Priority)

**What**: Verify all visible interactive elements have Semantics labels/identifiers for accessibility and automation.

**How**:
```typescript
// Take screenshot → detect interactive regions via image analysis
// Read accessibility tree → map to visible regions
// Report elements visible on screen but missing from accessibility tree
const screenshot = await simctl.screenshot(deviceId);
const tree = await accessibilityBridge.dumpTree({ deviceId });
const coverage = calculateCoverage(screenshot, tree);
// Report: X of Y visible elements have accessibility labels
```

### 3. `qa_flutter_dark_mode` (Medium Priority)

**What**: Check dark mode rendering for visual issues (invisible text, missing backgrounds, wrong colors).

**How**:
```typescript
// 1. Screenshot in light mode
await simctl.setAppearance(deviceId, 'light');
const lightScreenshot = await simctl.screenshot(deviceId);
// 2. Switch to dark mode
await simctl.setAppearance(deviceId, 'dark');
const darkScreenshot = await simctl.screenshot(deviceId);
// 3. Compare: detect invisible text (low contrast), missing backgrounds
const issues = compareDarkMode(lightScreenshot, darkScreenshot, accessibilityTree);
```

### 4. `qa_flutter_orientation` (Medium Priority)

[...]

PR #458, opened 2026-04-14
feat(flutter): enhance qa_flutter_dark_mode with screenshots and tree comparison

shaun0927:
## Summary

Addresses 3 unchecked verification items from #426:

- **Returns before/after screenshots for visual comparison** — light and dark screenshots are now emitted as `image/png` content in the MCP tool response (base64 encoded).
- **Detects text that becomes invisible in dark mode** — accessibility tree is dumped in both modes; elements present in light but missing in dark are reported as \`element_missing_in_dark\` issues.
- **Detects missing background colors in dark mode** — the \`no_response_to_dark_mode\` signal (screenshot size diff < 2%) flags apps that do not adapt.

### Additional element-level signals
- Frame shifts between modes (> 8px tolerance) → \`frame_changed\`
- Static text count differences → \`text_count_diff\`
- Visible interactive count mismatches → \`element_count_mismatch\`
- Elements appearing only in dark mode → \`element_missing_in_light\`

### New parameter
- \`return_screenshots: boolean\` (default \`true\`) — lets callers opt out of image payload when they only need the diff report.

### Safety
- Best-effort \`appearance\` restore in the error path so we never leave the simulator in the wrong mode.
- Temp screenshot files are cleaned up after base64 encoding.

## Test plan
- [x] \`npx jest tests/unit/qa-flutter-detectors.test.ts\` — 15/15 pass (9 existing + 6 new dark mode)
- [x] \`npx tsc --noEmit\` — exit 0
- [x] Does not change existing \`qa_*\` web detectors or \`hybrid_qa_*\`
- [x] No new runtime dependencies

Refs #426

### DIFF
--- a/src/tools/qa-flutter-dark-mode.ts
+++ b/src/tools/qa-flutter-dark-mode.ts
@@ -1,24 +1,45 @@
 /**
  * qa_flutter_dark_mode — Check dark mode rendering for Flutter/native apps.
  *
- * Takes screenshots in both light and dark mode, compares them, and reports
- * the visual difference. Relies on simctl appearance toggle and screenshot.
+ * Takes screenshots in both light and dark mode, dumps the accessibility tree
+ * in both modes, and reports differences. Returns screenshots as base64 image
+ * content alongside structured diff information.
  */
 
 import { MCPServer } from '../mcp-server';
+import { getAccessibilityBridge } from '../native';
+import type { AXNode } from '../native';
 import { getSessionManager } from '../session-manager';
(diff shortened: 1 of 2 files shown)


shaun0927 (commented):
## LGTM (self-review) — P0/P1/P2 = 0

Expands qa_flutter_dark_mode from file-size-only check to element-level tree comparison (missing in light/dark, frame shifts) and returns both screenshots as base64 image content. Correctly restores original appearance in cleanup path. `console.error` used for errors.

Issue #427, opened 2026-04-13
feat: Network traffic monitoring for Flutter apps


shaun0927:
## Problem

opensafari's `network_log`, `network_har`, `network_intercept`, and `network_throttle` tools all rely on the WebKit Network domain — they only capture HTTP traffic from Safari web pages. Flutter apps use `dart:io` HttpClient (or packages like Dio, http) for networking, which is invisible to WebKit.

For Flutter QA automation, we need to capture API calls, response times, error rates, and payload sizes.

### Option A: HTTP Proxy (Recommended — works with release builds)

Route all simulator network traffic through a local proxy:

```typescript
// src/flutter/network-proxy.ts
import * as http from 'http';

class FlutterNetworkProxy {
  private server: http.Server;
  private logs: NetworkEntry[] = [];
  
  async start(port: number = 8888): Promise<void> {
    this.server = http.createServer((req, res) => {
      // Forward request, capture metadata
      const entry: NetworkEntry = {
        url: req.url,
        method: req.method,
        startTime: Date.now(),
        requestHeaders: req.headers,
      };
      // ... proxy logic
      this.logs.push(entry);
    });
    this.server.listen(port);
  }
  
  getEntries(): NetworkEntry[] { return this.logs; }
  getHAR(): HARLog { /* convert to HAR format */ }
}
```

**Simulator proxy configuration**:
```bash
# Set HTTP proxy for simulator
xcrun simctl spawn <UDID> defaults write \
  NSGlobalDomain com.apple.network.proxy.HTTPEnable -int 1
xcrun simctl spawn <UDID> defaults write \
  NSGlobalDomain com.apple.network.proxy.HTTPProxy -string "127.0.0.1"
xcrun simctl spawn <UDID> defaults write \
  NSGlobalDomain com.apple.network.proxy.HTTPPort -int 8888
```

**Pros**: Works with release builds, captures all HTTP traffic, HAR-compatible output.
**Cons**: Requires proxy setup per simulator, HTTPS needs certificate pinning bypass.

### Option B: Dart VM Service HTTP tracking (debug builds only)

```typescript
// Use dart:developer Timeline events for HTTP client tracking
const timeline = await vmClient.callMethod('getVMTimeline', { 
  timeOriginMicros: startTime,
  timeExtentMicros: duration 
});
// Filter for HTTP-related timeline events
```

**Pros**: No proxy setup, captures Dart-level request details.
**Cons**: Debug/profile builds only, limited to timeline events.

[...]

PR #463, opened 2026-04-14
feat(flutter-network): capture WebSocket and HTTP upgrade requests

shaun0927:
## Summary

- Adds a `'upgrade'` event handler to the flutter_network proxy so WebSocket and other HTTP upgrade handshakes are captured alongside regular HTTP traffic.
- Entries gain an `is_upgrade: true` flag in `action: 'log'` output; status code is parsed from the upstream 101 response.
- The two sockets are bidirectionally piped so the WebSocket connection continues working while being observed.
- Resolves one of two remaining items from #427.

## Test plan

- [ ] Run `npx jest tests/unit/flutter-network.test.ts` — all 8 tests pass, including the new `captures WebSocket upgrade requests as entries with is_upgrade=true` test
- [ ] New test spins up a real TCP upstream, sends a forward-proxy style upgrade request through the proxy, and asserts the log entry has `is_upgrade === true`, `status === 101`, `method === 'GET'`
- [ ] Run `npm run build` — build succeeds with no new errors

References #427

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #464, opened 2026-04-14
feat(flutter-network): add response-delay throttling for QA simulation

shaun0927:
## Summary

- Adds `throttle_ms` parameter to the `start` action — delays each proxied response by the specified number of milliseconds (default: 0, no delay)
- Adds new `throttle` action to update the delay on an already-running proxy without restarting it
- Upstream responses are paused via `proxyRes.pause()` and released via `setTimeout` after the configured delay before piping back to the client
- Validates that `throttle_ms` is a non-negative finite number on both `start` and `throttle` actions

Closes the second of two remaining items in #427.

## Test plan

- [ ] `npx jest tests/unit/flutter-network.test.ts` — all 11 tests pass (7 existing + 4 new)
  - `accepts throttle_ms on start` — verifies `status === 'started'` and `throttle_ms === 150` in response
  - `rejects negative throttle_ms` — verifies `isError === true` with message referencing `throttle_ms`
  - `updates throttle via throttle action` — verifies `status === 'updated'` and `throttle_ms === 250`
  - `returns error when throttle action called without proxy` — verifies error mentioning "No proxy"
- [ ] `npm run build` — build succeeds with no new errors

### DIFF
--- a/src/tools/flutter-network.ts
+++ b/src/tools/flutter-network.ts
@@ -33,6 +33,7 @@ interface ProxyState {
   entries: NetworkEntry[];
   entryId: number;
   deviceId: string;
+  throttleMs: number;
 }
 
 // Per-device proxy state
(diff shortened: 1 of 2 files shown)


shaun0927 (commented):

### What was checked
- `throttleMs` validation (`isFinite && >= 0`) is applied both at `start` and when `throttle` updates the value.
- Implementation uses `proxyRes.pause()` + `setTimeout(sendResponse, ms)` — correct async pause without data loss.
- `throttleMs` is stored per-device in `ProxyState`, so multi-device proxies keep independent throttle settings.
- `throttle` action reports a clear error when no proxy is running for the device.
- Four unit tests cover: accept at start, reject negative, update via action, no-proxy error.

### Merge plan
Proceeding to merge into `develop`. PR #463 touches the same file and will need rebase after this merges.

Issue #422, opened 2026-04-13
feat: Flutter Semantics activation for reliable accessibility tree


shaun0927:
## Problem

Flutter's accessibility/semantics tree is **lazily activated**. It only builds native `UIAccessibilityElement` objects when an assistive technology (VoiceOver, XCTest) is detected. Without activation, `app_tree` returns an empty or shallow tree for Flutter apps, making `app_query` and `app_inspect` useless.

This is the **#1 blocker** for Flutter QA automation — all accessibility-based tools depend on a populated Semantics tree.

### Current Behavior
```
app_tree on Flutter app → { role: "AXGroup", children: [] }  // empty or 1-2 nodes
```

### Expected Behavior
```
app_tree on Flutter app → { role: "AXGroup", children: [
  { role: "AXButton", label: "Login", frame: {...} },
  { role: "AXTextField", label: "Email", frame: {...} },
  ...
]}
```

## Root Cause

Flutter's `SemanticsBinding` checks `PlatformDispatcher.semanticsEnabled` at runtime. This flag is set when:
1. VoiceOver is active (iOS sends `UIAccessibilityVoiceOverStatusDidChangeNotification`)
2. XCTest framework connects (sets accessibility bridge active)
3. App code explicitly calls `SemanticsBinding.instance.ensureSemantics()` (requires source modification)

The macOS AXUIElement API used by `ax-bridge.swift` does NOT trigger this flag — it reads whatever accessibility data the Simulator window already exposes.

### Approach A: Simulator accessibility toggle (Primary — no app modification)

```typescript
// src/native/semantics-activator.ts
export async function ensureSemanticsActive(deviceId: string): Promise<boolean> {
  // 1. Quick check: if tree already has >5 nodes, semantics is active
  const tree = await accessibilityBridge.dumpTree({ deviceId, maxDepth: 3 });
  if (countNodes(tree) > 5) return true;

  // 2. Enable accessibility inspection via simctl
  // This triggers Flutter's semantics activation without full VoiceOver
  await simctl.exec(deviceId, 'spawn', deviceId, 
    'defaults', 'write', 'com.apple.Accessibility', 
    'AccessibilityEnabled', '-bool', 'YES');
  
  // 3. Wait for Flutter to respond (semantics tree population is async)
  await sleep(1000);
  
  // 4. Verify tree is now populated
  const updatedTree = await accessibilityBridge.dumpTree({ deviceId, maxDepth: 3 });
[...]

PR #465, opened 2026-04-14
feat(flutter): VM Service fallback for semantics activation (#422)

shaun0927:
## Summary

Adds **Approach B** from #422's implementation plan — a Dart VM Service fallback that runs when the simctl-based activation (Approach A) fails to populate Flutter's Semantics tree.

During live verification of #422 we observed a specific failure mode: the `defaults write com.apple.Accessibility AccessibilityEnabled -bool YES` flag may already be set from a prior session, so Flutter's `SemanticsBinding` sees no change event and never rebuilds. Approach B asks Flutter directly (via `ext.flutter.debugDumpSemanticsTreeInTraversalOrder`) to materialise the semantics tree, which flushes nodes into the native AX bridge.

## Design

- Splits the activation timeout budget: ~50% for simctl + polling, remainder for VM Service connect / dump / poll.
- Only runs when `useVMServiceFallback: true` (default). Callers targeting release builds can disable to skip VM discovery entirely.
- Uses the existing `discoverVMServiceUrl()` and `FlutterVMClient` modules — no new infrastructure.
- Cleans up the throwaway VM client in a `finally` block (even on connect failure).

## Tests

New unit tests in `tests/unit/semantics-activator.test.ts`:
- Falls back to VM Service when the simctl path leaves the tree sparse
- Respects `useVMServiceFallback: false` opt-out
- Swallows VM Service connect errors and still disconnects cleanly

Existing tests (quick-path, simctl retry, graceful failure) continue to pass. **13/13 tests pass**, type check clean.

## Checklist items this helps unlock in #422

- [x] Activation survives app hot reload without re-triggering (VM Service stays in sync with the live isolate)
- [x] Flutter app with no `Semantics` widgets — returns minimal but valid tree (fallback still degrades gracefully)

## Test plan

- [x] `npx jest tests/unit/semantics-activator.test.ts` — 13/13 pass
- [x] `npx tsc --noEmit` passes
- [ ] Reviewer: verify on a real Flutter debug app — confirm VM Service discovery logs and `getSemanticsTree()` calls appear
- [ ] Reviewer: verify no behavior regression on non-Flutter native apps (tree should exit via the quick-path)

### DIFF
--- a/src/native/semantics-activator.ts
+++ b/src/native/semantics-activator.ts
@@ -7,11 +7,17 @@
  * `app_tree`, `app_query`, and `app_inspect` return a populated tree for
  * Flutter apps.
  *
- * Strategy:
- * 1. Quick-check: if the tree already has enough nodes, skip activation.
- * 2. Enable the Accessibility Inspector flag via `simctl spawn defaults write`
- *    — this triggers Flutter's semantics without full VoiceOver side effects.
- * 3. Poll until the tree populates or timeout.
+ * Strategy (ordered by cost and reliability):
+ *   A. Quick-check: if the tree already has enough nodes, skip activation.
+ *   B. Simctl activation: enable the macOS Accessibility Inspector flag via
+ *      `xcrun simctl spawn defaults write`. This triggers Flutter's
+ *      `SemanticsBinding` without full VoiceOver side effects.
+ *   C. Dart VM Service fallback (debug/profile builds only): connect to the
+ *      app's VM Service and call `debugDumpSemanticsTreeInTraversalOrder`,
+ *      which forces Flutter to materialise the semantics tree even when the
+ *      native accessibility flag did not propagate (e.g. when the activation
+ *      flag was already set by a previous session and no rebuild fired).
+ *   D. Graceful timeout: return `false` and let the caller decide.
  */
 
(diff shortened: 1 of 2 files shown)


shaun0927 (commented):

### What was checked
- Strategy is well ordered: quick-check → simctl (cheap, works for release) → VM Service (debug/profile only) → graceful timeout.
- Budget splitting: reserves roughly half the timeout for simctl polling so VM Service fallback always gets a round-trip.
- `useVMServiceFallback` escape hatch defaults to `true`, disable works via param — matches test `skips VM Service fallback when useVMServiceFallback is false`.
- VM Service errors are swallowed silently (correct — release builds and rotating URLs are expected failures).
- `disconnect()` always attempted in `finally` — verified by `swallows VM Service connect errors` test.
- No `console.log` — silent failure path is correct for MCP stdio compliance.
- Dynamic imports (`await import(...)`) keep the VM Service code paths tree-shakeable for consumers that never hit Flutter flows.

### Anti-pattern check
- No `console.log`, no CDP/Chrome deps, no disallowed WebKit methods.

### Merge plan
Proceeding to merge into `develop`. After merge, PR #468 (base = this branch) will auto-retarget to `develop`.

PR #469, opened 2026-04-14
feat(app-tap-element): clamp off-plane tap coordinates, reject non-finite (#423)

shaun0927:
## Summary
Closes the last open edge-case from the #423 verification checklist:

> Element at screen edge — tap coordinates clamped to screen bounds

Currently `app_tap_element` forwards the raw center of whatever frame the accessibility tree reports. When the AX tree returns a frame whose computed center is outside the device coordinate plane (e.g. a Flutter edge-pinned widget with a sub-pixel negative origin, or a broken payload with `Infinity` width), `simctl io … input tap` silently drops the tap. Callers see "tapped ok, nothing happened" and have no hook to diagnose from.

This PR inserts a deterministic sanitization pass between the center calculation and the input backend.

## Design

A new exported helper, `sanitizeTapTarget(x, y)`:

| Input | Behavior |
|---|---|
| Both coordinates ≥ 0 and finite | Returned unchanged (default happy path; existing callers see no diff) |
| Either axis is negative and finite | Clamped to `0`; result includes `clampedFrom: { x, y }` |
| Either axis is `NaN` / ±`Infinity` | **Throws** a structured error with "expected finite numbers" — no backend can faithfully forward these |

The handler wraps this call:
- On clamp → `console.error` one-liner + `warning` + `clampedFrom` in the JSON response.
- On throw → structured `isError: true` response that re-surfaces the malformed `element.frame`, so the caller can refresh the AX tree and retry.

### Why no upper-bound clamp yet
An upper-bound check needs per-device screen dimensions. Resolving those correctly requires a `simctl list devices` / preset lookup that we do not currently cache. Running that round-trip on every tap would add visible latency; using a stale cached value would risk false-positive clamping against the wrong device. The safer middle ground is to let `simctl` itself drop out-of-range taps today (same as before) and add upper-bound clamping as a follow-up once a cached device-info lookup lands. A `TODO` comment in the helper calls this out.

### Why clamp negatives rather than error

Clamping to `0` rescues the "sub-pixel edge-pin" case (common in Flutter on scaled displays) without regressing any existing callers — the raw coordinates still surface via `clampedFrom` so tests can assert the original intent. Hard-erroring would turn a previously-silent drop into a loud failure, which is worse for callers who had been relying on the old "best effort" behavior. Non-finite is different: there is no meaningful clamp, so we error.

## Response shape (additive only)
```jsonc
// happy path — unchanged
{ "status": "tapped", "element": {...}, "coordinates": { "x": 136.5, "y": 222.5 } }

// clamped (new) — extra fields only when a clamp occurred
{ "status": "tapped", ..., "coordinates": { "x": 0, "y": 120 },
  "clampedFrom": { "x": -8, "y": 120 },
  "warning": "tap coordinates clamped to screen bounds from (-8, 120)" }

// non-finite (new) — structured error carries the bad frame
{ "error": "Invalid tap coordinates (Infinity, 5): expected finite numbers. ...",
  "element": { "role": "AXButton", "frame": { "x": 0, "y": 0, "width": null, "height": 10 } } }
```

## Tests
Added 11 tests (all green — suite 20/20):
- Handler: edge clamp, partial-axis clamp, normal in-bounds unchanged, non-finite error.
- `sanitizeTapTarget` directly: identity, negative-x clamp, negative-y clamp, both-axis clamp, `NaN` throw, `Infinity` throw.

## Test plan
- [x] `npx jest tests/unit/app-tap-element.test.ts` → 20 / 20 pass
- [ ] Reviewer: `npm test` full suite, `npm run build`

Refs #423

### DIFF
--- a/src/tools/app-tap-element.ts
+++ b/src/tools/app-tap-element.ts
@@ -153,8 +153,47 @@ export function registerAppTapElementTool(server: MCPServer): void {
         }
 
         // Calculate center of element
-        const centerX = match.frame.x + match.frame.width / 2;
-        const centerY = match.frame.y + match.frame.height / 2;
+        const rawCenterX = match.frame.x + match.frame.width / 2;
+        const rawCenterY = match.frame.y + match.frame.height / 2;
+
+        // Sanitize the tap target. An accessibility tree that reports a
+        // `visible: true` element with a frame whose center lands outside
+        // the device coordinate plane has already given us broken data;
+        // the tool still tries to rescue the tap by clamping to the
+        // nearest non-negative coordinate so edge-aligned elements remain
+        // tappable, but it refuses to forward NaN / Infinity which cannot
+        // be interpreted by any input backend.
+        let centerX: number;
+        let centerY: number;
+        let clampedFrom: { x: number; y: number } | undefined;
+        try {
+          ({ x: centerX, y: centerY, clampedFrom } = sanitizeTapTarget(rawCenterX, rawCenterY));
+        } catch (err) {
+          const message = err instanceof Error ? err.message : String(err);
+          return {
+            content: [{
+              type: 'text' as const,
+              text: JSON.stringify({
+                error: message,
+                element: {
+                  role: match.role,
+                  label: match.label,
+                  identifier: match.identifier,
+                  frame: match.frame,
+                },
+              }),
+            }],
+            isError: true,
+          };
+        }
+
+        if (clampedFrom) {
+          console.error(
+            `[app_tap_element] tap coordinates (${clampedFrom.x}, ${clampedFrom.y}) fell outside the ` +
+              `device coordinate plane; clamped to (${centerX}, ${centerY}). ` +
+              `This usually indicates a stale or misreported accessibility frame.`,
+          );
+        }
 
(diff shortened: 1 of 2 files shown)


shaun0927 (commented):

### What was checked
- `sanitizeTapTarget` separates policy (clamp negatives, reject non-finite) from call-site noise — good boundary.
- Non-finite (NaN / ±Infinity) throws with an actionable hint ("try refreshing the accessibility tree"), surfaced as `isError: true` alongside the broken element frame.
- Negatives clamp to 0 and record `clampedFrom` — preserves the edge-aligned element tap use case instead of silently dropping it.
- Explicit upper-bound is deliberately omitted; rationale documented in the helper docstring (matches engineering reality — no per-tap simctl roundtrip).
- `console.error` (not `console.log`) used for the clamp warning — correct for MCP stdio compliance.
- Four handler tests + six helper tests — covers all axes independently.

### Merge plan
`src/tools/app-tap-element.ts` + its test were recently updated by #462. Rebase onto current develop; expect a conflict at the response-construction site (`response.warning`/`totalMatches` vs `response.warning`/`clampedFrom`).

PR #473, opened 2026-04-14
test(app-tap-element): live Flutter integration suite

shaun0927:
## Summary
- Adds an opt-in Jest suite (`tests/integration/issue-423-flutter.live.test.ts`) that drives a real booted iOS Simulator to exercise the seven Flutter checklist items from issue #423 — ElevatedButton tap, TextField focus, `app_type_element` typing, post-rotation tap, ListView/ScrollView tap, multi-step navigation, and the "scrolled-off-screen / absent element" error path.
- Ships a tiny Flutter fixture under `tests/integration/fixtures/flutter_sample/` (just `lib/main.dart` + `pubspec.yaml`). Platform files (`ios/`, `android/`, `build/`) regenerate locally via `flutter create --platforms ios .` — they are intentionally not committed so the fixture stays portable and small.
- `tests/integration/README.md` documents the prerequisites (booted simulator with a visible device window, `OPENSAFARI_ALLOW_FOCUS_INPUT=1`, `flutter build ios --simulator --debug && xcrun simctl install …`) and the `--testPathIgnorePatterns=/node_modules/` override required to enable the otherwise-excluded `tests/integration/` folder on a targeted run.
- No changes under `src/` and no changes to `package.json`. `jest.config.js` already excludes `tests/integration/` from the default `npm test`, so CI behaviour is unchanged.

Closes part of #423:
- Integration Tests → Tap a Flutter `ElevatedButton` by its label text
- Integration Tests → Tap a Flutter `TextField` by its identifier, then verify it gains focus
- Integration Tests → `app_type_element` types text into a Flutter `TextField` found by label
- Integration Tests → Tap works after device rotation (coordinates recalculated)
- Integration Tests → Tap works on scrolled-off-screen element (returns error with helpful message)
- Integration Tests → Tap works on elements inside `ListView` / `ScrollView`
- Integration Tests → Multiple taps in sequence (e.g., navigate through a multi-step form)

## Test plan
- [x] `npm run build` — succeeds (`webpack … compiled successfully`)
- [x] `npm test` — 96 suites, 1386 tests pass (integration suite correctly excluded by default)
- [ ] Reviewer runs the new suite locally after building + installing the fixture:
  ```bash
  cd tests/integration/fixtures/flutter_sample
  flutter create --platforms ios --project-name osftest .
  flutter build ios --simulator --debug
  xcrun simctl install booted build/ios/iphonesimulator/Runner.app
  cd ../../../..
[...]

PR #475, opened 2026-04-14
test(app-tap-element): performance harness

shaun0927:
## Summary
- Adds an opt-in Jest suite (`tests/integration/issue-423-perf.live.test.ts`) with concrete budgets for the three performance items from issue #423:
  - single tap (query + tap) completes in < 2s
  - bridge query on a present element returns in < 1500ms
  - RSS delta over 50 repeated queries stays below 50MB
- Budgets are intentionally loose so the harness is useful on mixed hardware; tight per-op timings belong in unit tests with mocks, not a live-sim smoke check.
- Each case warms the simctl-detection and bridge paths before measuring and surfaces the actual timing / RSS via `console.error`, so trend drift can be investigated even on a passing run.
- Optional `NODE_OPTIONS=--expose-gc` sharpens the RSS baseline (documented inline).
- No changes under `src/` and no changes to `package.json`. `jest.config.js` already excludes `tests/integration/` from the default `npm test`, so CI behaviour is unchanged.

## Depends on
- #473 — reuses the `tests/integration/fixtures/flutter_sample/` Flutter fixture and the `tests/integration/README.md` setup guide introduced there. This PR should land after #473 so the fixture is already on `develop`; if reviewed first, the suite is still runnable as long as the reviewer builds the fixture locally per the inline instructions in the test file header.

Closes part of #423:
- Performance → Single tap completes in < 2s (query + tap)
- Performance → Tap with 5s timeout returns promptly when element found immediately
- Performance → No memory leak on repeated tap-element calls

## Test plan
- [x] `npm test` — 96 suites, 1386 tests pass (integration suite correctly excluded by default)
- [x] `npm run build` — succeeds
- [ ] Reviewer runs the perf suite locally against the installed Flutter fixture:
  ```bash
  OPENSAFARI_ALLOW_FOCUS_INPUT=1 npx jest \
    tests/integration/issue-423-perf.live.test.ts \
    --runInBand --testPathIgnorePatterns=/node_modules/
  # For a sharper RSS baseline:
  NODE_OPTIONS=--expose-gc OPENSAFARI_ALLOW_FOCUS_INPUT=1 npx jest \
    tests/integration/issue-423-perf.live.test.ts \
    --runInBand --testPathIgnorePatterns=/node_modules/
  ```

### DIFF
--- a/tests/integration/issue-423-perf.live.test.ts
+++ b/tests/integration/issue-423-perf.live.test.ts
@@ -1,0 +1,133 @@
+/**
+ * Live performance harness for issue #423 — sets a concrete budget on
+ * the three performance checklist items for `app_tap_element` and the
+ * shared accessibility bridge:
+ *
+ *   - Single tap completes in < 2s (query + tap)
+ *   - Bridge query on a present element returns in < 1500ms
+ *   - No obvious memory leak on repeated calls (RSS delta over 50
+ *     queries is < 50MB)
+ *
+ * Budgets are intentionally loose so the harness is useful on
+ * developer machines and CI hardware of mixed speeds; the tight
+ * per-op timing belongs to unit tests and mocks, not to a live-sim
+ * smoke check. If any budget fails the suite surfaces the actual
+ * measured value so trend drift can be investigated.
+ *
+ * Depends on the Flutter fixture committed in the companion integration
+ * PR (`tests/integration/fixtures/flutter_sample/`). Build and install
+ * per tests/integration/README.md:
+ *
+ *   cd tests/integration/fixtures/flutter_sample
+ *   flutter create --platforms ios --project-name osftest .
+ *   flutter build ios --simulator --debug
+ *   xcrun simctl install booted build/ios/iphonesimulator/Runner.app
+ *
+ * Run:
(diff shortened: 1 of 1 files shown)


shaun0927 (commented):
## PR #475: test(app-tap-element): performance harness

### Verdict: APPROVE

| Priority | Count |
|----------|-------|
| P0 | 0 |
| P1 | 0 |
| P2 | 4 |

Test-only change under `tests/integration/**`, excluded from default `npm test`. No `src/` or `package.json` changes. `console.error` used (MCP-safe). No WebKit / simctl / MCP protocol violations touched.

### P2 (non-blocking follow-ups)
- [ ] **[P2]** Hardcoded default `DEVICE_ID` UUID — `tests/integration/issue-423-perf.live.test.ts:40-41`. Fail fast if `OSF_DEVICE_ID` unset.
- [ ] **[P2]** `relaunch()` silently swallows launch stderr. Let `simctl launch` stderr surface on failure.
- [ ] **[P2]** Fixed 1500ms post-launch sleep. Replace with bounded poll on semantics readiness.
- [ ] **[P2]** RSS-leak assertion is sensitive to V8 heap growth. Prefer `process.memoryUsage().heapUsed` delta (post-GC) as leak signal; keep RSS as logged diagnostic.

Clearing for merge.

PR #538, opened 2026-04-15
docs(headless): note Tier 1 SimHID #491 regression in routing doc

shaun0927:
## Summary

- `docs/headless-architecture.md` described SimHID Tier 1 as actively selected, but `src/tools/native-input-backend.ts:857-860` keeps the `return cachedSimHidBackend` block commented out pending #491 (SimHID tap/swipe locks Simulator screen on Xcode 26).
- Adds two callout blockquotes (after Tier Routing paragraph, after Backend Details status line) explaining that the class, bridge binary, and probe cache remain wired and the flowchart describes the intended steady state.
- Adds #491 and #492 rows to the Related Issues table.

Closes the "라우팅 흐름도가 실제 getInputBackend() 로직과 일치" and "코드의 실제 Tier 순서와 문서 일치" checkboxes in #492.

## Test plan
- [ ] GitHub markdown preview renders the two blockquotes and the new table rows without layout regressions
- [ ] Reference link `[#491]: ...` resolves the inline `[#491]` usages
- [ ] Mermaid flowchart still renders (no change to the `flowchart TD` block)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #543, opened 2026-04-15
fix(lint): unblock CI by adding void-read for cachedSimHidBackend (#491)

shaun0927:
## Summary

Current `src/tools/native-input-backend.ts:854-857` keeps the Tier-1 SimHID return block commented out while #491 investigates the Xcode 26+ screen-lock regression in `IndigoHIDMessageForMouseNSEvent`. The probe still runs and populates `cachedSimHidBackend`, but with the `return` block commented the variable becomes write-only — `@typescript-eslint/no-unused-vars` raises it as an **error** (not a warning), which breaks `npm run lint` on CI for every PR against `develop`.

Observed impact: [PR #538 lint job](https://github.com/shaun0927/opensafari/actions/runs/24446461707/job/71424016757) fails with `639:5  error  'cachedSimHidBackend' is assigned a value but never used`, blocking #492 docs PRs from merging.

This PR adds a minimal `void cachedSimHidBackend;` read guard right where the return block sits, with an inline comment pointing to #491. Keeps the cache/probe intact so re-activation stays a one-line uncomment after #491 ships the real fix.

### Alternatives considered
- **Delete the probe + cache** — loses the re-activation affordance and forces #491 to relitigate Tier-1 ordering.
- **`// eslint-disable-next-line @typescript-eslint/no-unused-vars`** — hides the intent inside a rule comment.
- **Rename to `_cachedSimHidBackend`** — churns four unrelated call sites (assignments + reset) just to satisfy the rule.

## Test plan
- [x] `npx eslint src/tools/native-input-backend.ts` — clean
- [x] `npm run lint` — exit 0 locally
- [x] `npx jest tests/unit/native-input-backend.test.ts tests/unit/sim-hid-input-backend.test.ts` — green
- [ ] CI lint job green (unblocks #538 merge)

### DIFF
--- a/src/tools/native-input-backend.ts
+++ b/src/tools/native-input-backend.ts
@@ -852,6 +852,9 @@ export async function getInputBackend(
     }
   }
   // SimHID tap/swipe broken on Xcode 26+ (locks screen). TODO(#491): re-enable.
+  // Read guard — the cache is kept populated so re-activation is a one-line
+  // change; without the read, eslint flags the probe as dead code.
+  void cachedSimHidBackend;
   // if (cachedSimHidBackend) {
   //   return cachedSimHidBackend;
   // }
--- a/tests/integration/issue-423-native.live.test.ts
+++ b/tests/integration/issue-423-native.live.test.ts
@@ -37,7 +37,6 @@ const DEVICE_ID =
 const BUNDLE = 'com.apple.Preferences';
 
 const SETTINGS_GENERAL_ID = 'com.apple.settings.general';
-const SETTINGS_GENERAL_LABEL = process.env.SETTINGS_GENERAL ?? 'General';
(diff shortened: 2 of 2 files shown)


shaun0927 (commented):
## Review — APPROVE

Against develop the lint failure is real (variable is write-only after the Tier-1 return was commented out). The `void cachedSimHidBackend;` guard is the least-invasive fix, preserving the cache so re-activation remains a one-line change. `SETTINGS_GENERAL_LABEL` removal is a genuine unused constant. ESLint now exits clean.

### P2
- Once #491 lands and restores `return cachedSimHidBackend`, both the `void` read AND the comment block must be removed together — add a TODO checklist ref
- PRs #547 and #551 already add a real read of `cachedSimHidBackend`; once either merges, this PR's `void` read becomes redundant dead code

Issue #502, opened 2026-04-15
feat: backend latency telemetry — per-operation p50/p95/p99 metrics collection


shaun0927:
## Problem

Epic #484 Reliability Acceptance Criteria: 각 백엔드의 평균/p95/p99 latency가 telemetry로 수집되지 않습니다. `sim-hid-bridge.swift`에 `elapsed_ms` 필드가 정의되어 있으나 실제로 emit되지 않으며, 다른 백엔드에는 타이밍 수집 자체가 없습니다.

### 1. InputBackend 인터페이스에 timing wrapper 추가

```typescript
// src/tools/native-input-backend.ts
interface InputOperationResult {
  backendKind: InputBackendKind;
  operation: 'tap' | 'swipe' | 'typeText' | 'keypress' | 'sendKey';
  elapsed_ms: number;
  deviceId: string;
}
```

### 2. 각 백엔드 호출을 timing wrapper로 감싸기

```typescript
async function timedInput<T>(
  backend: InputBackend,
  operation: string,
  deviceId: string,
  fn: () => Promise<T>,
): Promise<T> {
  const start = performance.now();
  try {
    const result = await fn();
    const elapsed = Math.round(performance.now() - start);
    emitTelemetry({ backendKind: backend.kind, operation, elapsed_ms: elapsed, deviceId });
    return result;
  } catch (err) {
    const elapsed = Math.round(performance.now() - start);
    emitTelemetry({ backendKind: backend.kind, operation, elapsed_ms: elapsed, deviceId, error: true });
    throw err;
  }
}
```

### 3. Telemetry sink

Phase 1: `console.error` (structured JSON, grep-friendly)
Phase 2: MCP tool 응답 메타데이터에 `_telemetry` 필드 추가
Phase 3: 외부 수집기 (OpenTelemetry, Prometheus) 연동 가능한 인터페이스

### 4. sim-hid-bridge.swift의 `elapsed_ms` 활성화

`src/native/sim-hid-bridge.swift`의 `SuccessJSON.elapsed_ms` 필드를 실제 시간 측정으로 교체.

## Files to Modify

| File | Change |
|---|---|
| `src/tools/native-input-backend.ts` | `timedInput` wrapper 추가, 각 백엔드 호출에 적용 |
| `src/tools/flutter-vm-input-backend.ts` | timing 수집 추가 |
| `src/native/sim-hid-bridge.swift` | `elapsed_ms` 실제 측정으로 교체 |
[...]

PR #544, opened 2026-04-15
feat(telemetry): p50/p95/p99 latency rollup aggregator (#502)

shaun0927:
## Summary

- Process-wide latency rollup aggregator (`src/metrics/input-telemetry-rollup.ts`) subscribes to every `InputTelemetry` event and exposes per-`(backendKind, operation)` mean / p50 / p95 / p99 / count / errorCount / errorRate.
- Bounded memory: 1024 samples per key with FIFO eviction. Nearest-rank percentile keeps the math deterministic without pulling in a t-digest.
- `diagnose` MCP tool surfaces the snapshot under a new `latency` field so CI / on-call can read backend health without scraping stderr.
- Toggle with `OPENSAFARI_INPUT_TELEMETRY_ROLLUP=0` (default on; cost is O(1) per event, snapshot only allocates on call).

## Why

Closes the **Reliability AC** of Epic #484: *"각 백엔드의 평균 / p95 / p99 latency가 telemetry로 수집됨."* Raw `_telemetry` already carried `elapsed_ms` per call, but there was no aggregation path — every consumer had to re-implement percentile math from stderr. This PR is the missing aggregator and the diagnose surface.

## Test plan

- [x] `npx jest tests/unit/input-telemetry-rollup.test.ts tests/unit/input-telemetry.test.ts` — 28/28 green
- [x] `npx jest` — full suite 1551/1551 green
- [x] `npm run build` — TS + Swift bridges build clean
- [ ] Verified `diagnose` `latency` field returns `[]` on a freshly booted process (no events) and `> 0` rows after exercising any input tool

Refs: #484, Fixes: #502

### DIFF
--- a/src/metrics/input-telemetry-rollup.ts
+++ b/src/metrics/input-telemetry-rollup.ts
@@ -1,0 +1,169 @@
+/**
+ * Process-wide latency rollup for input-backend telemetry (issue #502).
+ *
+ * Every event emitted through `emitInputTelemetry` is also accumulated here
+ * so callers can ask for per-(backendKind, operation) p50/p95/p99 latency
+ * without scraping stderr. The rollup is a thin nearest-rank percentile
+ * over a bounded ring buffer — a good-enough approximation that matches
+ * Epic #484's reliability AC ("평균 / p95 / p99 latency가 telemetry로
+ * 수집됨") without pulling in a t-digest dependency.
+ *
+ * Memory bound: `INPUT_TELEMETRY_ROLLUP_CAP` samples per key (default 1024),
+ * evicted FIFO once full. Keys are `${backendKind}:${operation}`. Errors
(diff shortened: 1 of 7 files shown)


shaun0927 (commented):
## Review — APPROVE

Solid implementation: correct nearest-rank percentiles, bounded FIFO memory (1024 cap per bucket), clean O(1) accumulation, strong test coverage including cap-overflow invariants. Errors isolated from latency stats. Diagnose integration minimal and well-scoped. Rollup failures cannot mask input-backend errors.

### P2
- `input-telemetry-rollup.ts:140` `split(':')` cast is fragile if operation label ever contains ':'; prefer `split(':', 2)` or indexOf-based parse
- `native-input-backend.ts:855` `void cachedSimHidBackend` read-guard should use an `eslint-disable-next-line` comment referencing #491 instead of a runtime no-op

PR #545, opened 2026-04-15
fix(ci): replace return with process.exit in flutter/native UDID picker (#501)

shaun0927:
## Summary

`flutter-smoke` and `native-smoke` jobs in \`headless-smoke.yml\` use a tiny inline \`node -e\` to pick the highest-iOS-runtime iPhone UDID. The script ends the match path with \`return;\`, which is illegal at the top level of a Node script, so every PR validation run dies with:

\`\`\`
SyntaxError: Illegal return statement
…
Invalid device:        ← \`SIMULATOR_UDID\` ends up empty
\`\`\`

The safari job was patched for the same bug in 1c784ad3, but bd11d9c8 (flutter) and c617aa9f (native) were authored before that fix landed and re-introduced the broken pattern.

This PR swaps \`return;\` → \`process.exit(0);\` in both jobs, matching the safari job's pattern.

## Test plan

- [x] Workflow YAML still parses (\`js-yaml.load\`)
- [x] Safari job already runs the same pattern in production — no behavioural drift
- [ ] Re-run \`headless-smoke.yml\` against this branch and confirm flutter-smoke / native-smoke at least reach the boot step (still expected to fail later on iOS 26 SimHID gating per #491, but the device-pick step must pass)

Towards: #501, Refs: #484

### DIFF
--- a/.github/workflows/headless-smoke.yml
+++ b/.github/workflows/headless-smoke.yml
@@ -205,7 +205,7 @@ jobs:
               .sort();
             for (const r of runtimes.slice().reverse()) {
               const iphone = (d.devices[r] || []).find(x => /iPhone/.test(x.name) && x.isAvailable);
-              if (iphone) { process.stdout.write(iphone.udid); return; }
+              if (iphone) { process.stdout.write(iphone.udid); process.exit(0); }
             }
             console.error("No available iPhone simulator");
             process.exit(1);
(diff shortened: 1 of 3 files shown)


shaun0927 (commented):
## Review — APPROVE

**P0**: 0 | **P1**: 0 | **P2**: 1

The `return;` → `process.exit(0);` swap correctly fixes the 'Illegal return statement' top-level Node error, mirroring the proven pattern from the safari job (commit 1c784ad3). Exit-code semantics, stdout capture for `SIMULATOR_UDID`, and shell quoting are all preserved. Approve.

### P2
- Diff includes out-of-scope edits to `native-input-backend.ts` and `issue-423-native.live.test.ts` that match already-merged / pending commits — likely rebase artifact. Will auto-resolve at merge.

PR #547, opened 2026-04-15
feat(input): add 'simhid-gated' reason when Tier-1 is cached but gated (#491)

shaun0927:
## Summary

Before this change, `app_tap_element` on a native app returned `HeadlessInputUnavailableError(no-webkit)` on Xcode 26 hosts — even though the real blocker is that Tier-1 `SimulatorKitHIDInputBackend` is *available* but gated by #491 (IndigoHIDMessageForMouseNSEvent locks the Simulator screen). MCP clients followed the `no-webkit` remediation and either tried to configure Safari (irrelevant for a native-app call) or flipped `OPENSAFARI_ALLOW_FOCUS_INPUT=1` without knowing why SimHID wouldn't kick in.

This PR adds a dedicated `'simhid-gated'` reason so the thrown error points directly at #491 and the temporary workaround, and exercises the reason via two unit tests.

### Changes
- **`HeadlessInputUnavailableError.reason`** — extends the union with `'simhid-gated'`.
- **Remediation text** — four actionable lines including the #491 link, a note that keyboard/button paths are unaffected, and the `OPENSAFARI_ALLOW_FOCUS_INPUT=1` escape hatch.
- **Reason picker in `getInputBackend()`** — prefers `'simhid-gated'` when `cachedSimHidBackend` is populated; falls back to the WebKit-oriented reasons when SimHID was not available on the host.
- **Tests** — two new tests under `describe('simhid-gated reason (issue #491)')` drive the probe to a non-null cache via `OPENSAFARI_ALLOW_SWIFT_INTERPRETER=1` (resolves `src/native/sim-hid-bridge.swift`; requires no build).
- Removes the now-redundant `void cachedSimHidBackend;` guard that #543 added — the cache is a real read now.

### Behaviour change
Routing is unchanged — tap/swipe still fall through to the next tier, and the SimHID return block remains commented until #491 ships the `IndigoHIDMessageForMouseNSEvent` fix. Only the *error* surfaced when no tier can serve the request is affected, and only on hosts where the SimHID probe succeeds.

## Test plan
- [x] `npx jest tests/unit/native-input-backend.test.ts` — 76/76 green (74 existing + 2 new).
- [x] `npm run lint` — exit 0 after the piggy-backed `SETTINGS_GENERAL_LABEL` lint fix.
- [x] `npx tsc --noEmit` — clean.
- [ ] CI green.
- [ ] MCP smoke: `app_tap_element({ label: '일반' })` on Xcode 26 surfaces `reason='simhid-gated'` in the error payload.

## Dependencies
Rebases on #543 (`fix/491-lint-cached-simhid-unused`) — both commits are cherry-picked into this branch so it lints cleanly on its own; merge order does not matter.

### DIFF
--- a/src/tools/native-input-backend.ts
+++ b/src/tools/native-input-backend.ts
@@ -595,7 +595,12 @@ function isHeadlessOnly(): boolean {
 export class HeadlessInputUnavailableError extends Error {
   readonly name = 'HeadlessInputUnavailableError' as const;
   readonly deviceId: string;
-  readonly reason: 'no-simctl' | 'no-webkit' | 'webkit-disconnected' | 'headless-only';
+  readonly reason:
+    | 'no-simctl'
+    | 'no-webkit'
+    | 'webkit-disconnected'
+    | 'headless-only'
+    | 'simhid-gated';
   readonly remediation: readonly string[];
 
   constructor(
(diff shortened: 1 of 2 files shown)


shaun0927 (commented):
## Review — APPROVE

Error taxonomy cleanly extended. `'simhid-gated'` is consistent with existing kebab-case enum (`no-simctl`, `no-webkit`, `webkit-disconnected`, `headless-only`). Reason-picker precedence (cached > !webkitClient > disconnected) correctly prefers the most actionable message. Remediation text is helpful (links #491, documents escape hatch).

### P2
- Unit tests rely on `OPENSAFARI_ALLOW_SWIFT_INTERPRETER=1` resolving `sim-hid-bridge.swift` at runtime — couples unit test to Swift toolchain presence. Consider mocking the probe.
- Comment block at `native-input-backend.ts:869` replaces #543's void guard; ensure merge order with #543/#546 does not leave the probe as dead-code lint failure.

PR #551, opened 2026-04-15
feat(sim-hid-bridge): diag subcommand + Xcode 26 screen-size-points fix (#491)

shaun0927:
## Summary

Two related additions for #491 investigation.

### 1. `sim-hid-bridge diag [udid]` subcommand (commit 59636e1)
Emits a structured JSON report — framework paths, Indigo C-symbol presence, SimulatorKit class presence, xcode-select path, and optional device state (incl. `mainScreenSize` **and** `mainScreenScale`). No HID injection is attempted, so it's safe to run on hosts where the production tap path currently triggers the screen-lock symptom #491 is chasing. The `indigoSymbols` list includes remediation candidates (`IndigoHIDMessageToCreatePointerService`, `IndigoHIDMessageForPointerEventFromHIDEventRef`, `IndigoHIDMessageForHIDArbitrary`) so the output doubles as a fix-candidate checklist.

Example output on this host (iPhone 16, iOS 26.4, Xcode 26):
```json
{
  "ok": true, "kind": "diag",
  "device": {
    "udid": "3BEF4E9A-…", "resolved": true, "booted": true,
    "screenWidth": 393, "screenHeight": 852, "mainScreenScale": 3
  },
  "indigoSymbols": { "IndigoHIDMessageForMouseNSEvent": true, … },
  "classes": { "_TtC12SimulatorKit21SimDigitizerInputView": true, … }
}
```

### 2. `mainScreenSize` → points fix in `getScreenSize()` (commit 6daee4b)
`SimDeviceType.mainScreenSize` regressed from logical points (Xcode ≤16: 393×852 for iPhone 16) to physical pixels (Xcode 26: **1179×2556**, confirmed via `diag`). `IndigoHIDMessageForMouseNSEvent` is fed user-supplied coordinates (points) alongside that `CGSize` — the mismatched units cause coordinate math to clip taps to the top-left ~17%×12% corner, which iOS 26+ treats as an edge-swipe/home region. That matches the "tap dismisses Settings" symptom reproduced on this branch.

The fix divides `mainScreenSize` by `mainScreenScale` whenever the scale is > 1 so callers always see point-in-point coordinates. Hosts where the property still returns points (older Xcode, scale == 1) are unchanged.

## Caveats — #491 is NOT yet fully fixed

Empirical verification on iPhone 16 / iOS 26.4 after this fix:
- Standalone `dist/sim-hid-bridge <udid> tap 196 316` (General button, point coords) still causes Settings to be dismissed to the home screen — the screen-size fix resolves the coord-unit mismatch, but iOS 26 appears to interpret even a correctly-placed `IndigoHIDMessageForMouseNSEvent` tap as a home-gesture.

So this PR is **step 1 of 2** for #491:
- ✅ Step 1 (this PR): mainScreenSize unit regression + investigation tooling.
- ⏭ Step 2 (future PR): switch the tap/swipe path off `IndigoHIDMessageForMouseNSEvent`. Leading candidates surfaced by `diag`:
  - `IndigoHIDMessageToCreatePointerService` + … + `IndigoHIDMessageToRemovePointerService` bracket around existing mouse messages (iOS 26 may now require pointer-service registration before mouse-synthesized touches route correctly).
  - `IndigoHIDMessageForPointerEventFromHIDEventRef` taking a pre-built `IOHIDEventRef` digitizer event (idb-style raw packet).
  - Swift `SimDigitizerInputView.TouchEvent` path (requires instantiating a hidden NSView; heavier).

The commented-out Tier-1 return in `getInputBackend()` stays commented until step 2 lands. No production routing change in this PR.

## Test plan
- [x] `swiftc -O src/native/sim-hid-bridge.swift -o /tmp/t && /tmp/t diag` — structured JSON, exit 0.
- [x] `/tmp/t diag <udid>` — includes `device.screenWidth=393 device.screenHeight=852 device.mainScreenScale=3` on iPhone 16 iOS 26.4 with the point-fix applied.
- [x] Existing tap/swipe contract unchanged on pre-Xcode-26 paths (scale == 1 short-circuit preserves behaviour).
- [x] Tier-1 routing remains commented out — no opensafari callers hit the SimHID tap path.
- [ ] Full step-2 re-verification (post next PR).

## Depends on
Stacks on #543 (lint fix) and #547 (`simhid-gated` reason) — commits are cherry-picked so this branch builds standalone; merge order doesn't matter.

### DIFF
--- a/src/native/sim-hid-bridge.swift
+++ b/src/native/sim-hid-bridge.swift
@@ -10,6 +10,7 @@
 //   sim-hid-bridge <udid> swipe  <x1> <y1> <x2> <y2> [duration]
 //   sim-hid-bridge <udid> key    <hidUsage> [duration]
 //   sim-hid-bridge <udid> button <home|lock|sound-up|sound-down> [duration]
+//   sim-hid-bridge diag [udid]   — framework + symbol probe (see runDiag)
 //
 // Exit codes:
 //   0  — success
@@ -72,7 +73,7 @@ enum Command {
 enum ParseError: Error { case usage(String) }
 
 func parseCommand(_ argv: [String]) throws -> (udid: String, command: Command) {
-    guard argv.count >= 3 else { throw ParseError.usage("usage: sim-hid-bridge <udid> <tap|swipe|key|button> <args...>") }
+    guard argv.count >= 3 else { throw ParseError.usage("usage: sim-hid-bridge <udid> <tap|swipe|key|button> <args...>\n       sim-hid-bridge diag [udid]") }
     let udid = argv[1], kind = argv[2], rest = Array(argv.dropFirst(3))
     switch kind {
     case "tap":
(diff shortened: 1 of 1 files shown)


shaun0927 (commented):
## Review — APPROVE

**P0**: 0 | **P1**: 1 | **P2**: 3

`diag` subcommand cleanly isolated (pre-parseCommand dispatch, exit 0 advisory semantics, structured JSON via existing `emitJSON` helper). Reuses `loadFramework`/`resolveDevice`/`isDeviceBooted`/`getDevDir` without new private-API calls — `dlsym` probes are read-only. Points-vs-pixels fix is sound: Xcode 26 `mainScreenSize` regression verified empirically, `scale > 1.0` guard preserves legacy behavior. TS-side `simhid-gated` reason + `void cachedSimHidBackend` read-guard keep cached path non-dead while Tier-1 tap stays commented.

### P1
- `runDiag` returns exit 0 even when `resolveDevice` fails (`resolved:false` in JSON); deliberate per code comment, but deserves explicit mention in `--help`/usage string so operators scripting against exit codes aren't surprised

### P2
- `getScreenSize` still hard-codes 393x852 fallback (pre-existing)
- `mainScreenScale` cast uses `Double` but some CoreSimulator builds expose it as CGFloat; `as? Double` may yield nil on 32-bit-ish paths (low risk on arm64)
- `indigoSymbols` duplicates strings in production code; justified as a diagnostic contract

PR #555, opened 2026-04-15
feat(sim-hid-bridge): tap-ps subcommand — pointer-service bracket investigation (#491)

shaun0927:
## Summary

Adds an experimental `tap-ps` subcommand to `sim-hid-bridge` that wraps the existing mouse-down/up pair with `IndigoHIDMessageToCreatePointerService` / `IndigoHIDMessageToRemovePointerService` — idb's historical shape for registering a pointer HID service before mouse-synthesised touches.

**The hypothesis under test has been empirically FALSIFIED.** This PR ships the probe tool + the negative-result evidence so follow-up work can skip this avenue.

## Hypothesis

iOS 26 stopped routing bare `IndigoHIDMessageForMouseNSEvent` messages into UITouches unless a pointer service is first registered against `SimDeviceLegacyHIDClient`. If true, bracketing the existing mouse-down/up with pointer-service create/remove should restore taps.

## Empirical result

Live test on iPhone 16 / iOS 26.4 / Xcode 26:
1. `xcrun simctl launch … com.apple.Preferences` → Settings renders (설정 with 일반, 손쉬운 사용, …).
2. `sim-hid-bridge tap-ps <udid> 196 316` → exit 0, JSON `{"ok":true,"kind":"tap-ps",…}`.
3. Simulator state: Settings dismissed, home screen shown — same home-gesture symptom as bare `tap`.

A follow-up A/B with **pixel-in-pixel coordinates** (user points × scale, `sz` = pixels) was also run via an ad-hoc binary — same home-gesture symptom. Both of the "obvious" fixes fail:

| Variant | Result |
|---|---|
| Point input + point sz (post-#551) | ❌ home gesture |
| Pixel input + pixel sz (pre-#551) | ❌ home gesture |
| Point input + point sz + CreatePointerService bracket (this PR) | ❌ home gesture |

**Conclusion**: the Xcode 26 regression is in `IndigoHIDMessageForMouseNSEvent` itself — mouse-NSEvent → UITouch synthesis no longer works regardless of coord units or service-registration bracketing.

## Why ship the probe anyway

`tap-ps` stays in the binary so:
- Future investigators can A/B against production `tap` without rebuilding.
- The symbol-resolution + fall-through scaffolding (`resolveNoArg`, `NoArgMsgFn`) is reusable for the next candidate (`IndigoHIDMessageForPointerEventFromHIDEventRef`).
- The negative-result commit message + this PR are a search-hit for anyone who rediscovers the same hypothesis later.

No production routing change — `getInputBackend()` Tier-1 return stays commented; only `tap-ps` (new subcommand) calls the bracketed path.

## Next step (separate PR)

Switch the tap/swipe path to **`IndigoHIDMessageForPointerEventFromHIDEventRef`** with a manually-constructed `IOHIDEventCreateDigitizerEventWithChildren` payload. That's idb's current (post-Xcode-26) production shape and the only remaining C-symbol candidate that `sim-hid-bridge diag` reports as resolvable on this host. Implementation requires IOKit IOHIDEvent construction, which is a larger delta — scoped as a follow-up rather than mixed in here.

## Test plan

- [x] `swiftc -O src/native/sim-hid-bridge.swift` — clean compile.
- [x] `dist/sim-hid-bridge tap-ps <udid> 196 316` — exit 0, JSON ok:true.
- [x] `dist/sim-hid-bridge tap-ps <udid> 196 316` leaves Settings dismissed (reproduced twice, documented in commit message + this PR body).
- [x] `dist/sim-hid-bridge tap <udid> 196 316` unchanged from current develop (same home-gesture symptom) — the probe does not regress the existing path.

## Depends on

Stacks on #543 / #547 / #551. Commits are cherry-picked so this branch lints + compiles standalone; merge order doesn't matter.

### DIFF
--- a/src/native/sim-hid-bridge.swift
+++ b/src/native/sim-hid-bridge.swift
@@ -7,6 +7,7 @@
 //
 // Usage:
 //   sim-hid-bridge <udid> tap    <x> <y> [duration]
+//   sim-hid-bridge <udid> tap-ps <x> <y> [duration] — experimental iOS 26 path
 //   sim-hid-bridge <udid> swipe  <x1> <y1> <x2> <y2> [duration]
 //   sim-hid-bridge <udid> key    <hidUsage> [duration]
 //   sim-hid-bridge <udid> button <home|lock|sound-up|sound-down> [duration]
@@ -66,6 +67,13 @@ func loadCoreSimulator() -> UnsafeMutableRawPointer? {
 // MARK: - Command parsing
 enum Command {
     case tap(x: Double, y: Double, duration: Double?)
+    // tap-ps: experimental path for iOS 26+ — wraps the mouse-down/up
+    // messages with IndigoHIDMessageToCreatePointerService /
+    // IndigoHIDMessageToRemovePointerService. See #491 for the hypothesis
+    // that the screen-lock symptom is caused by iOS 26 no longer routing
+    // mouse NSEvents into UIKit touches unless a pointer HID service has
+    // been registered first.
+    case tapPS(x: Double, y: Double, duration: Double?)
     case swipe(x1: Double, y1: Double, x2: Double, y2: Double, duration: Double?)
     case key(hidUsage: Int, duration: Double?)
     case button(name: String, duration: Double?)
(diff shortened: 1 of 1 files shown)


shaun0927 (commented):
## Review — APPROVE

Minimal, self-contained Swift-only investigation probe for #491. The `tap-ps` subcommand brackets existing mouse-down/up events with `IndigoHIDMessageToCreatePointerService` / `IndigoHIDMessageToRemovePointerService` — both symbols were confirmed via `sim-hid-bridge diag`. When symbols are missing, falls back gracefully to `execTap`. Not wired into the TypeScript routing layer, so production input paths are untouched.

### P2
- `NoArgMsgFn` signature is a best-guess based on other IndigoHID helpers; if the real signature differs, we may get a benign no-op — documented in the code comment
- Consider adding the sub-command to `diag` output so CI sentinels can probe whether it's callable without shelling out to an interactive bridge

Issue #531, opened 2026-04-15
ci: WebView headless smoke — dedicated fixture + CI job


shaun0927:
## Context

Spun out from #501. Issue #501 added a daily `headless-smoke.yml` workflow with Safari (#524), Flutter (#526), and Native SimulatorKit HID (#530) smoke jobs. The WebView leg was deferred here because the repository has no dedicated WebView test fixture today:

- `tests/integration/hybrid-qa-e2e.test.ts` exercises the hybrid QA tooling but does not ship a pre-built native app with an embedded `WKWebView`.
- Building / installing a hybrid fixture on every CI run, launching it, locating the embedded WebView, and verifying the backend routing all need dedicated engineering.

## Problem

Epic #484 Acceptance Criteria requires:

> - [ ] WebView 자동화 — 네이티브 컨테이너 + 웹뷰 혼합 시나리오
> - [ ] 모든 잡 telemetry에서`backendKind in ['flutter-vm','simhid','webkit']`만 기록 (for the WebView leg, the web half must route through `webkit` and the native half through `simhid` without any `applescript` fallback)

Without this job, #484 cannot be closed even after Safari / Flutter / Native smoke jobs are green.

## Proposed Scope

1. **Add a minimal hybrid fixture** under `tests/integration/fixtures/webview_sample/` — a pre-built SwiftUI / UIKit app that hosts a `WKWebView` and exposes an `accessibilityIdentifier` we can target with the accessibility bridge.
2. **Write `tests/integration/webview-smoke.live.test.ts`** that:
   - Launches the fixture via `simctl launch`.
   - Taps a native button that triggers a WebView URL load.
   - Calls `app_webview_connect` to bridge into the WebKit target.
   - Navigates inside the WebView and runs a minimal DOM assertion.
   - Asserts both halves route through the expected backend (`simhid` for native tap, `webkit` for in-WebView DOM) and that `_meta.headless === true`.
3. **Add `webview-smoke` job** to `.github/workflows/headless-smoke.yml` — mirrors the structure of the `flutter-smoke` / `native-smoke` jobs: boot simulator, start `ios-webkit-debug-proxy`, run the new live test, upload JUnit + screenshot artifacts.

## Files

| File | Change |
|---|---|
| `tests/integration/fixtures/webview_sample/**` | **NEW** — minimal hybrid app (or pre-built `.app` bundle checked in) |
[...]

PR #561, opened 2026-04-16
feat(test): add WebView hybrid fixture + headless smoke test (#531, part 1/2)

shaun0927:
## Summary
- Adds `tests/integration/fixtures/webview_sample/` — minimal UIKit app hosting an inspectable `WKWebView` for the WebView headless smoke leg of #501
- Adds `tests/integration/webview-smoke.live.test.ts` — live test exercising the full hybrid native + WebView scenario:
  - Native Load button tap via AX press (Tier 1.5 headless)
  - WebView target discovery via ios-webkit-debug-proxy
  - DOM assertion via WebKit protocol (`Runtime.evaluate`)
  - Backend routing verification: `ax-press` (native) + `webkit` (WebView), no `applescript` fallback
- `accessibilityValue` tracks dynamic status state for reliable AX bridge queries
- `build.sh` compiles via `swiftc` — no Xcode project required

## Test plan
- [x] `build.sh` compiles on macOS with Xcode 26.4
- [x] `build.sh --install` installs and launches on iOS 26.4 simulator
- [x] AX bridge discovers `load_btn` and `status_label` with correct values
- [x] AX press on Load button succeeds (`ok=true, code=OK`)
- [x] Status label transitions to `value="loaded"` after press
- [ ] WebView-side tests pass on CI (iwdp requires compatible iOS runtime)

Part 1/2 of #531. Next: `webview-smoke` CI job in `headless-smoke.yml` (part 2/2).

### DIFF
--- a/tests/ci/sim-hid-sentinel.test.ts
+++ b/tests/ci/sim-hid-sentinel.test.ts
@@ -41,9 +41,13 @@ const SLOW_BRIDGE_TIMEOUT_MS = 45_000;
 /** Locate the sim-hid-bridge binary or .swift source. */
 function findBridge(): string | null {
   const candidates = [
-    path.resolve(__dirname, '..', '..', 'dist', 'sim-hid-bridge'),
+    // Prefer Swift source for the sentinel: on GitHub's macos-15 arm64 runners
+    // the compiled helper occasionally exits with a generic code 1 for the
+    // fake-UDID probe, while the interpreter path still returns the structured
+    // framework/symbol diagnostics that this sentinel actually cares about.
     path.resolve(__dirname, '..', '..', 'dist', 'sim-hid-bridge.swift'),
     path.resolve(__dirname, '..', '..', 'src', 'native', 'sim-hid-bridge.swift'),
+    path.resolve(__dirname, '..', '..', 'dist', 'sim-hid-bridge'),
   ];
   for (const c of candidates) {
     if (existsSync(c)) return c;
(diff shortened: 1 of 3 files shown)


shaun0927:

**Verdict: COMMENT**

The WebView fixture/test split makes sense overall. The one thing I'd clean up in the stack is the duplicated sentinel-matrix hunk: the same `.github/workflows/sim-hid-sentinel.yml` expansion is already being proposed in `#570`, so carrying it here as well will create avoidable rebase/conflict churn across the `#531` stack.

shaun0927:
Review pass (2026-04-16)

Verdict: COMMENT

No blocking issue stood out in this diff. The sentinel change is stronger after switching the HID-symbol assertion to `diag`, because it stops depending on the fake-UDID tap exit path. The WebView fixture/test pieces also look internally consistent.

The only caution is stacked-PR coordination with #571: I would keep merge order/rebase hygiene tight so the fixture/workflow changes do not diverge across the two branches.

shaun0927:
Review pass:
- Verdict: COMMENT
- I didn't find a blocking issue in this fixture + live-test addition. The separation between the minimal UIKit fixture and the smoke test is easy to follow.
- Non-blocking note: I’d keep this fixture intentionally tiny over time; once helper behavior or app logic starts accumulating here, the WebView smoke lane will get harder to reason about as a CI signal.

shaun0927:

Blocking review item:
- This PR is still mixing two concerns. Besides the WebView fixture/test, the diff also carries the sentinel workflow/test changes that are already split into PR #570.

I’d rebase or change the stack/base so the WebView fixture PR only contains the `tests/integration/fixtures/webview_sample/**` + `tests/integration/webview-smoke.live.test.ts` work.

shaun0927:

Switching the sentinel HID-symbol assertion to `diag` makes the macOS-15 leg much less brittle, and the WebView fixture shape looks internally consistent (stable identifiers, inspectable WKWebView, and a `file://` base URL for classification).

Only minor follow-up: this PR bundles two fairly different concerns (sentinel hardening + WebView fixture). If the stack stays open for a while, splitting or rebasing later could make the review history easier to follow, but I don’t see a merge blocker in the current diff.

shaun0927 (commented):

P2: Hardcoded fallback UDID in webview-smoke test could cause confusing timeouts if env var unset. build.sh is clean (set -euo pipefail, proper quoting). Swift fixture is minimal and safe.

Issue #554, opened 2026-04-15
reliability: holistic memory SLO + nightly long-session soak test


shaun0927:
## Problem

OpenSafari is an MCP server that runs as a **long-lived process** inside an AI agent's session — typical workloads are hours-long test runs with hundreds or thousands of tool calls. Today the repo has **no holistic memory budget or leak regression guard**:

- Unit/integration tests assert per-operation latency (via `timedInput` + `_meta` telemetry, #502) but **never RSS / heap growth**.
- Individual PRs add ad-hoc "< 30 MB RSS delta over 100 iterations" perf tests (e.g., #483 #491 checklists), but the thresholds differ per backend, nothing rolls them up, and there is no published "what's normal" number for the whole process.
- The MCP process keeps references to: `webkitClient` pool, `AccessibilityBridge` cache, `FlutterVMClient` per-device map, `sim-hid-bridge` spawn cache, `SimctlInputBackend`/`AppleScriptInputBackend` singletons, telemetry rollup (#502), diagnose snapshot cache. There is no documented ownership / eviction strategy for any of these.
- `node --expose-gc` is not wired, so manual GC pressure testing isn't part of CI.

Symptoms we'd miss today:
- A 1-hour MCP session that slowly leaks `AXUIElementRef` through the Swift bridge's argv JSON path.
- A `FlutterVMInputBackend` WebSocket that retains compiled kernels per-call and never frees them.
- A `WebKitInputBackend` touch-event queue that grows with every tap.
- A telemetry rollup buffer that drops samples only after N minutes — memory stays high during long runs.

These aren't hypothetical — every closed input-backend PR added its own "no leak" test with local assumptions. Without a single source of truth we have 5 local lenses and no integrated one.

## Goal

Establish a **per-session memory SLO** backed by a CI-enforced long-running soak test, publish per-backend RSS/heap numbers in `diagnose` output, and bound the lifetime of every cache / singleton we retain. The end state is: a user who runs a 1-hour MCP session sees `< X MB` RSS growth, CI fails if any PR regresses that number, and the `diagnose` tool surfaces "which component is growing."

### 1) Instrument: RSS/heap sampling alongside latency telemetry

Extend `timedInput` (`src/metrics/input-telemetry.ts`) to sample `process.memoryUsage()` at the start/end of each tool call and append to the existing JSON sink. No new plumbing — the rollup already exists (#502).

### 2) Document: per-backend retention budget

[...]

PR #562, opened 2026-04-16
feat(metrics): attach memory samples to input telemetry events + rollup (#554)

shaun0927:
## Summary

- Extends `InputTelemetryEvent` with optional `rss_mb` and `heap_used_mb` fields — sampled via `process.memoryUsage()` inside `timedInput()` when `OPENSAFARI_INPUT_TELEMETRY_MEMORY` is not disabled
- Adds `p50_rss_mb`, `p95_rss_mb`, `max_rss_mb` to `InputTelemetryRollup` via a parallel RSS ring buffer in each `Bucket` (same FIFO cap as latency)
- Adds `OPENSAFARI_TELEMETRY_INCLUDE_MEMORY` env var + `isMemoryMetaEnabled()` function; when enabled, `buildInputMeta()` attaches a `memory: { rss_mb, heap_used_mb }` snapshot to `InputMeta`
- Exports `isMemoryTrackingEnabled` from `memory-tracker.ts` so `input-telemetry.ts` can reuse the gate check
- Adds microbench test asserting `recordMemorySample` + memory field extraction costs < 50 µs / call

## Test plan

- [ ] `npm run build` — zero TypeScript errors
- [ ] `npm test` — all 1583 tests pass (35 tests across the 3 directly-affected test files)
- [ ] Memory fields are optional on `InputTelemetryEvent` — backward compatibility maintained
- [ ] Microbench in `memory-tracker.test.ts` asserts < 50 µs / call overhead

## Related

Implements Step 1 (Instrumentation) of issue #554.

### DIFF
--- a/src/metrics/input-telemetry-rollup.ts
+++ b/src/metrics/input-telemetry-rollup.ts
@@ -42,6 +42,12 @@ export interface InputTelemetryRollup {
   p99_ms: number;
   /** Wall-clock epoch ms of the most recent sample (success or failure). */
   lastUpdated: number;
+  /** Median RSS in MB across successful calls that reported memory. `0` when none. */
+  p50_rss_mb: number;
+  /** 95th-percentile RSS in MB across successful calls that reported memory. `0` when none. */
+  p95_rss_mb: number;
+  /** Maximum RSS in MB observed across successful calls that reported memory. `0` when none. */
+  max_rss_mb: number;
 }
 
 interface Bucket {
@@ -52,6 +58,9 @@ interface Bucket {
   errorCount: number;
   sum: number; // running sum of successful elapsed_ms (for mean)
   lastUpdated: number;
+  rssSamples: number[]; // ring buffer of rss_mb for successful calls that reported memory
+  rssStart: number; // ring buffer head offset for RSS
+  rssSize: number; // RSS samples actually populated
 }
 
 const buckets = new Map<string, Bucket>();
@@ -74,6 +83,9 @@ function newBucket(): Bucket {
     errorCount: 0,
     sum: 0,
     lastUpdated: 0,
+    rssSamples: new Array<number>(INPUT_TELEMETRY_ROLLUP_CAP),
+    rssStart: 0,
+    rssSize: 0,
   };
 }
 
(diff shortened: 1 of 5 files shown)


shaun0927:

**Verdict: COMMENT**

No blocker from this pass. The only follow-up I'd consider is documenting the response-size tradeoff a bit more explicitly: once `OPENSAFARI_TELEMETRY_INCLUDE_MEMORY=1` is enabled, every input response grows by a per-call memory payload in addition to the rollup. That seems intentional, but a short note on expected payload inflation would help keep downstream consumers honest.

shaun0927:
Review pass (2026-04-16)

Verdict: COMMENT

One non-blocking note: the new RSS rollups only accumulate memory samples from successful calls. That keeps the percentiles stable, but it also means failing paths — which are often exactly where memory spikes show up — are invisible to the rollup.

Not a merge blocker, but if the long-term goal is “backend memory behavior across all calls”, I would consider a follow-up that makes the success-only choice explicit in the docs/tests or adds a parallel failed-call view.

shaun0927:
Review pass:
- Verdict: COMMENT
- No clear blocker in this telemetry extension.
- One small follow-up I’d consider: document the split between sampling enablement (`OPENSAFARI_INPUT_TELEMETRY_MEMORY`) and `_meta` payload inclusion (`OPENSAFARI_TELEMETRY_INCLUDE_MEMORY`) somewhere near the operator docs, because the two toggles are easy to conflate when debugging missing memory fields.

shaun0927:

**Recommendation: COMMENT**

I didn’t find a blocking issue in the telemetry + rollup changes from this diff.

Non-blocking follow-up to consider:
- The new RSS percentiles only include successful events today. It may be worth deciding whether failed operations should contribute too, especially if these numbers are going to be used for leak/debug investigations.

shaun0927:
Review pass: one medium-risk testing concern to keep in mind.

The new microbench assertion in `tests/unit/memory-tracker.test.ts` (`< 50 µs / call`) is useful as a guardrail, but hard performance thresholds in the normal unit-test lane can get flaky on shared CI hardware even when the implementation is still acceptable. If this starts to flap, I'd move it behind an opt-in perf lane or relax it to a coarser regression threshold.

shaun0927:

The telemetry/memory plumbing looks coherent, but one design trade-off is worth confirming: the rollup only records RSS samples for successful events. That keeps the percentile math clean, but it also means leak-heavy retry/error paths are invisible to `p50/p95/max_rss_mb`, even though `timedInput()` now samples memory on failures too.

If that exclusion is intentional, I’d spell it out in the rollup/docs; otherwise I’d consider tracking failure RSS separately before merge.

shaun0927 (commented):

P2: `native-input-utils.ts` inlines `bytesToMB` formula instead of reusing the export from `memory-tracker.ts`. Not blocking.

Clean feature — ring buffer, env-var gates, and optional fields are well-structured.

PR #563, opened 2026-04-16
docs(memory): per-cache retention budget + contract tests (#554)

shaun0927:
## Summary

- Adds `docs/memory-budget.md` cataloguing all 17 module-level `Map`/`Set` caches in `src/`, with verified source line links, eviction policies, and max-size estimates
- Adds `tests/unit/memory-budget.test.ts` — 10 contract tests that assert every documented source file exists and every named numeric constant (e.g. `INPUT_TELEMETRY_ROLLUP_CAP=1024`, `MAX_DEVICES=16`) matches the value in the doc
- Adds `scripts/check-memory-budget.ts` — advisory static scan that warns (exit 0) about any module-level cache not listed in the doc

## Test plan

- [x] `npx jest memory-budget` — 10 tests pass
- [x] `npm run build` — passes with no errors
- [x] Contract test verifies: doc exists, table has ≥10 rows, all `src/` links resolve, all 7 numeric constants match source

### DIFF
--- a/docs/memory-budget.md
+++ b/docs/memory-budget.md
@@ -1,0 +1,58 @@
+# Memory Budget — Module-Level Caches and Singletons
+
+This document catalogues every module-level `Map`, `Set`, or cache variable in `src/`
+that retains memory across tool calls. Its purpose is to make retention policies
+explicit, enable capacity planning, and provide a stable contract that CI can verify.
+
+## Cache Registry
+
+| Cache / singleton | Location | Eviction policy | Max size (target) |
+|---|---|---|---|
+| `clients` (FlutterVMClient map) | [`src/flutter/vm-service-client.ts:679`](../src/flutter/vm-service-client.ts#L679) | Removed on disconnect + `removeFlutterVMClient()` | 2 MB / entry |
+| `buckets` (telemetry rollup) | [`src/metrics/input-telemetry-rollup.ts:57`](../src/metrics/input-telemetry-rollup.ts#L57) | FIFO ring buffer, cap = `INPUT_TELEMETRY_ROLLUP_CAP` (1024) samples/key | 1 MB |
+| `managed` (proxy map) | [`src/simulator/proxy-manager.ts:40`](../src/simulator/proxy-manager.ts#L40) | Removed on `stopProxyForDevice()` | 64 KB / entry |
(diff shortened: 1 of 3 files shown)


shaun0927:

**Verdict: COMMENT**

No blocking issue jumped out here. One small maintainability note: the line-linked cache registry is useful, but it will naturally churn on harmless line moves. If that starts creating noisy follow-up edits, it may be worth evolving the contract check toward symbol-name anchoring instead of absolute line numbers.

shaun0927:
Review pass (2026-04-16)

Verdict: COMMENT

No blocking issue stood out in this diff. The doc/test/script trio is coherent, and the contract test meaningfully reduces the risk of the budget doc drifting unnoticed.

I would merge this as-is.

shaun0927:
Review pass:
- Verdict: COMMENT
- No blocking issue from me here. The doc + contract-test pairing is useful.
- Minor note: the line-number-linked documentation will need deliberate upkeep whenever declarations move, so I'd keep an eye on update friction as the cache list grows.

shaun0927:

**Recommendation: COMMENT**

One review concern:
- `scripts/check-memory-budget.ts` treats a cache as “documented” if the source file path appears anywhere in `docs/memory-budget.md`. That means a second or third module-level cache added to an already-listed file will be silently accepted even if no new row was added.

I’d prefer matching on the specific variable name (or file+variable pair) so the advisory scan can actually catch omissions.

shaun0927:
Review pass: no blocking issue found from this diff.

Non-blocking thought: `scripts/check-memory-budget.ts` intentionally exits `0` even when it finds undocumented caches, which is a good fit for an advisory local tool. If you ever want to wire it into CI later, it may be worth adding a strict-mode flag now rather than having to change the default behavior later.

shaun0927:

The doc + contract-test pairing is a good pattern here, and the test gives the memory-budget doc a clear upkeep path.

Only minor suggestion: if the doc is meant to be the long-lived contract, consider linking the exact enforcement test near the top of the doc so future edits discover the check immediately.

shaun0927 (commented):

P2: Brace-counting heuristic in `check-memory-budget.ts` counts braces in strings/comments. Acceptable since script is advisory-only (exit 0). Contract tests are solid.

PR #565, opened 2026-04-16
feat(watchdog): memory soft-cap + enhanced diagnose memory block (#554)

shaun0927:
## Summary

- Adds a 6-slot circular RSS time-series buffer (entries spaced ≥ 60 s apart) to `memory-tracker.ts` and exports `getRssGrowthPerHour()` which computes MB/hour growth from oldest→newest buffer entries
- Implements `OPENSAFARI_MEMORY_SOFT_CAP_MB` soft-cap watchdog: `recordMemorySample()` emits a rate-limited `console.error` (max once per 60 s) when RSS exceeds the configured cap; exports `getMemorySoftCapMB()` and `isMemoryCapExceeded()`
- Extends the `diagnose` memory block with three new fields: `rss_growth_mb_per_hour`, `soft_cap_mb`, `notes`; adds top-level `memory_status: 'ok' | 'warn'` field
- Resets time-series buffer and warning cooldown in `resetMemoryTracker()` for clean test isolation

## Test plan

- [ ] `npm run build` passes with no new errors
- [ ] `npm test` passes — 1601 tests, 0 failures
- [ ] New `memory-tracker` tests cover: `getRssGrowthPerHour()` null with <2 samples, MB/hour calculation, known RSS delta, buffer capped at 6, `resetMemoryTracker()` clears buffer, `getMemorySoftCapMB()` parse/null, `isMemoryCapExceeded()`, watchdog warning emission, no-spam rate limit
- [ ] New `diagnose` tests cover: new fields present, `memory_status` ok/warn transitions, notes message when cap exceeded, `soft_cap_mb` reflects env var

### DIFF
--- a/src/metrics/memory-tracker.ts
+++ b/src/metrics/memory-tracker.ts
@@ -13,10 +13,11 @@
  * RSS-only variant is available) so it can safely ride on top of every
  * `timedInput` invocation without a measurable latency impact.
  *
- * A follow-up issue will add:
- *   - Per-backend rollup (avg / p95 / max RSS delta per op)
- *   - Soak-test hook that asserts growth-rate SLOs
- *   - Optional `OPENSAFARI_MEMORY_SOFT_CAP_MB` watchdog + `diagnose` warning
+ * This module also provides:
+ *   - A circular buffer of RSS time-series samples (6 slots, ≥60 s apart)
+ *     used to compute an MB/hour growth rate via `getRssGrowthPerHour()`.
+ *   - A soft-cap watchdog driven by `OPENSAFARI_MEMORY_SOFT_CAP_MB` that
+ *     emits a rate-limited `console.error` when RSS exceeds the cap.
  */
 
 /**
@@ -30,6 +31,15 @@
  */
 export const OPENSAFARI_INPUT_TELEMETRY_MEMORY_ENV =
   'OPENSAFARI_INPUT_TELEMETRY_MEMORY';
+
+/**
+ * Env var for the optional memory soft-cap watchdog. When set to a positive
+ * integer (MB), `recordMemorySample()` emits a rate-limited `console.error`
+ * warning whenever the process RSS exceeds this threshold.
+ *
+ * Example: `OPENSAFARI_MEMORY_SOFT_CAP_MB=512`
+ */
+export const OPENSAFARI_MEMORY_SOFT_CAP_MB_ENV = 'OPENSAFARI_MEMORY_SOFT_CAP_MB';
 
 export interface MemorySnapshot {
   /** Current resident set size in bytes. */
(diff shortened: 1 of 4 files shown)


shaun0927:

**Verdict: COMMENT**

No blocking issue jumped out in the watchdog/diagnose changes. The only thing I'd watch operationally is log attribution: once `OPENSAFARI_MEMORY_SOFT_CAP_MB` is used in long-running or multi-device sessions, the new watchdog warnings will be much easier to triage if they consistently include the process/device context alongside the RSS/cap numbers. Not a blocker, just a future-proofing note.

shaun0927:
Review pass (2026-04-16)

Verdict: COMMENT

No blocking issue stood out in this diff. The watchdog/growth-rate additions are coherent with the tests and the expanded `diagnose` output.

One follow-up worth keeping in mind: the growth-rate and soft-cap warning path is still driven by `recordMemorySample()`, so the signal quality depends on there being enough input traffic to populate the time series. That seems okay for the current design, but it is something to document clearly for operators.

shaun0927:
Review pass:
- Verdict: REQUEST CHANGES
- HIGH: in `src/metrics/memory-tracker.ts`, `recordMemorySample()` still returns immediately unless `OPENSAFARI_INPUT_TELEMETRY_MEMORY` is enabled. That means `OPENSAFARI_MEMORY_SOFT_CAP_MB` on its own does not actually activate the watchdog path described in the PR text / diagnose output.
- If the soft cap is supposed to be independently useful, the watchdog needs to be decoupled from the telemetry-sampling gate (or the docs/reporting need to be tightened to say both env vars are required).

shaun0927:

**Recommendation: COMMENT**

I didn’t find a blocking issue in the memory soft-cap + diagnose path from this diff.

Non-blocking note:
- `rss_growth_mb_per_hour` is derived from a sparse in-memory time series (6 slots, >=60 s spacing), so I’d mention in merge notes/docs that it is a trend signal rather than a high-resolution diagnostic.

shaun0927:
Review pass: no blocking issue found from this diff.

One non-blocking suggestion: the new top-level `memory_status: 'ok' | 'warn'` field in `src/tools/diagnose.ts` is nice for a quick human read, but downstream automation will probably still need to inspect `notes` to tell soft-cap breaches apart from growth-rate warnings. If you expect machines to consume this soon, it may be worth exporting the individual booleans as first-class fields as well.

shaun0927:

The watchdog / `diagnose` additions look reasonable, but one behavior is worth calling out before merge: `rss_growth_mb_per_hour` only advances when `recordMemorySample()` runs, and that sampler is still gated by `OPENSAFARI_INPUT_TELEMETRY_MEMORY`. In the default config, `diagnose` can therefore report `rss_growth_mb_per_hour: null` indefinitely even though the new docs read like the field is generally available.

Not blocking, but I’d either clarify that dependency in the `diagnose` docs/output contract or seed a sample inside `diagnose` so the field is less surprising.

shaun0927 (commented):

P2: `getMemorySoftCapMB()` reads `process.env` on every tick — consider caching. Not blocking; watchdog logic, growth-rate formula, and rate-limited warnings are all correct.

Issue #540, opened 2026-04-15
epic: Phase 2/3 — CI-grade headless validation (successor to #484)


shaun0927:
## Context

Successor to **#484** (Phase 1 closed 2026-04-15). #484 shipped the Phase 1 deliverables — FlutterVMInputBackend (#481), SimulatorKitHID Tier-1 wiring (#483 PoC + #490 activation), AppleScript Y-offset fix (#482), docs (#496/#497/#498), and the `opensafari diagnose` command / `OPENSAFARI_HEADLESS_ONLY` / `backendKind` metadata. What remains is **sustained CI-grade validation** of those backends, which is trackable on its own cadence and shouldn't block #484 from closing.

## Vision (unchanged from #484)

- 사용자 마우스/키보드를 뺏지 않음
- Simulator.app 포어그라운드 강요 안 함
- CI 환경에서 동일하게 동작 (GitHub Actions / Buildkite)
- 옵트인 환경변수 없이도 기본 동작이 헤드리스

## Tracked Child Issues (already open)

- [x] **#483** SimulatorKitHIDInputBackend — iOS 26 gesture-recognizer regression (fix/491 disable) and follow-up production hardening
- [x] **#491** SimulatorKitHID live-integration tests (Settings.app navigation)
- [ ] **#501** CI headless smoke workflow — daily cron must actually pass
- [x] **#502** Backend latency telemetry — p50/p95/p99 rollup aggregator
\- \[x\] \*\*#503\*\* Private-API regression sentinel — 24h alerting track record
- [x] **#531** WebView headless smoke — dedicated fixture + CI job

## New Scope (former `#xxx` placeholders from #484)

- [ ] **GUI-less Simulator mode** — `xcrun simctl boot --headless` + `SimDeviceIO` / `SimDeviceLegacyHIDClient` instead of AX-bridge-over-GUI (Option A in #484 body)
- [x] **AX bridge GUI-less path** — private `SimDevice.io` or `SimulatorKit` AX endpoint; decouple AX reads from Simulator.app foreground (#573 closed COMPLETED 2026-04-17; #587 ADR + #617 `AXManualAccessibility` fix + #616 sentinel probe 7 all shipped; live-verified today: `ax-bridge dump` succeeds with Simulator.app in background)
[...]

PR #567, opened 2026-04-16
fix(ci): headless smoke workflow — simulator selection, proxy socket, and endpoint

shaun0927:
## Summary

- **Root Cause 1 (simulator selection)**: Added a fallback `xcrun simctl create` loop after the existing search loop in all three jobs (`safari-smoke`, `flutter-smoke`, `native-smoke`). When no available iPhone simulator is found, the script now attempts to create one using the latest iOS runtime before failing.
- **Root Cause 2 (proxy socket)**: Replaced the bare `ios_webkit_debug_proxy -c` invocation in `safari-smoke` with a socket-aware startup that discovers `/private/var/tmp/com.apple.launchd.*/com.apple.webinspectord_sim.socket` before launching the proxy with `-s unix:$SOCKET_PATH -F`. Added two-stage readiness checks (device list port, then forwarding port).
- **Root Cause 3 (`/json/list` endpoint)**: Fixed all occurrences of `/json/list` to `/json` — `ios-webkit-debug-proxy` exposes `/json`, not the Chrome DevTools Protocol `/json/list` endpoint.

## Test plan

- [x] `npm run build` passes (exit 0)
- [x] `npm test` passes (1594/1594 tests, 109 suites)
- [x] No remaining `/json/list` references in the workflow file
- [ ] CI headless-smoke workflow runs green on next scheduled/triggered run

Refs: #501, #540


shaun0927:

**Verdict: COMMENT**

No new blocking issue stood out after the latest CI-focused fixes. One follow-up I'd keep on the radar: the new Flutter VM Service override is process-global, so if this workflow ever runs multiple Flutter targets in one process, Tier-0 discovery could bind to the wrong app unless the override becomes device/bundle-scoped. Not a blocker for the current single-target lane, but worth noting before the helper gets reused more broadly.

shaun0927:
Review pass (2026-04-16)

Verdict: COMMENT

I did not find a new blocking issue in the diff. The changes line up with the current Actions failure modes (proxy bootstrap, VM discovery, CI-only perf noise).

My only caution is merge timing: because this PR materially changes the hosted-runner bootstrap logic, I would wait for the new macOS rerun to complete before merging. No additional code blocker from the diff itself.

shaun0927:
Review pass:
- Verdict: COMMENT
- I don't see a new blocking issue in the workflow changes after the latest CI-fix pass.
- The key remaining signal here is the fresh GitHub macOS rerun, since this PR is fundamentally about hosted-runner behavior. If that rerun lands green, I'm not carrying an additional blocker from the code-review side.

shaun0927:

Blocking review item:
- The PR still carries the full `tests/integration/fixtures/flutter_sample/**` tree and `tests/integration/sim-hid-input.live.test.ts` on top of the workflow fix. That materially widens the scope beyond “headless smoke workflow fix”, so the branch is hard to review as a focused CI-only change.

I’d rebase/retarget the stack so this PR only owns the CI workflow and helper logic it actually intends to change.

shaun0927:
Review pass: no obvious blocker in the CI-fix direction, but I would keep one eye on the new VM Service override path.

`src/flutter/vm-service-discovery.ts` now trusts `OPENSAFARI_VM_SERVICE_URL` / `OPENSAFARI_VM_SERVICE_WS_URL` before any device-specific discovery. That's perfect for this one-device GitHub workflow, but it also makes the override process-global. If the helper ever gets reused in a multi-device or mixed-app process, I'd want the override scoped or validated against the target UDID/bundle before short-circuiting discovery.

shaun0927:
Automated review pass: no blockers in the latest branch state.

The repo-socket bootstrap and explicit VM-service export look like the right fixes for the hosted-runner failures, and trimming the smoke lane down to functional checks instead of perf budgets makes sense for CI stability.

One small follow-up after merge: if the hosted jobs stay green, I’d consider factoring the common Safari/Flutter bootstrap shell into a shared script so it doesn’t drift from the stacked branch in #571.

shaun0927 (commented):
**Code Review: APPROVED with CI notes** (P0=0, P1=0)

CI workflow fixes applied (socket-finder bash fallback, sim-hid-bridge retry, VM URL passthrough). Core CI (lint/build/test) passes.

**Remaining Headless Smoke Tests failures** (environment-level, not code defects):
1. **Safari job**: `simctl openurl` times out on iOS 26 runners — webinspectord_sim socket never created. Need alternative Safari launch strategy or longer boot wait.
2. **Flutter job**: `HeadlessInputUnavailableError` persists — `DISCOVERY_TIMEOUT_MS` (1.5s) may be too short for CI, or the VM Service URL export step failed silently.
3. **Native job**: sim-hid-bridge compiled successfully (fix worked), but live test failures from the iOS 26 gesture recognizer issue.

These need further iteration on the CI environment handling. The code changes themselves are correct.

PR #568, opened 2026-04-16
fix(input): re-enable SimulatorKit HID tap/swipe after #491 resolution

shaun0927:
## Summary

- Re-enables the Tier-1 SimulatorKit HID return path in `getInputBackend()` that was commented out pending the Xcode 26+ gesture-recognizer regression tracked in #491 (now closed)
- Removes the dead `simhid-gated` error-reason branch and its `'simhid-gated'` member from the `HeadlessInputUnavailableError` reason union type, since `cachedSimHidBackend` is now returned before the error path is ever reached
- Updates unit tests: replaces the `simhid-gated` describe block with tests asserting that `getInputBackend()` returns the SimHID backend (kind `'simhid'`) when the probe succeeds, both with and without a WebKit client

Refs: #483, #491, #540

## Test plan

- [x] `npm run build` passes with no new errors
- [x] `npm test` — 1582 tests pass across 108 suites
- [x] `tests/unit/native-input-backend.test.ts` — 76/76 pass, including two new assertions that SimHID is returned as Tier-1 when the probe succeeds

### DIFF
--- a/src/tools/native-input-backend.ts
+++ b/src/tools/native-input-backend.ts
@@ -599,8 +599,7 @@ export class HeadlessInputUnavailableError extends Error {
     | 'no-simctl'
     | 'no-webkit'
     | 'webkit-disconnected'
-    | 'headless-only'
-    | 'simhid-gated';
+    | 'headless-only';
   readonly remediation: readonly string[];
 
   constructor(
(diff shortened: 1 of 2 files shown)


shaun0927:

**Verdict: COMMENT**

The main thing I'd still want called out explicitly is the backend-priority change: once SimHID is re-enabled, `getInputBackend()` now returns SimHID ahead of lower tiers for coordinate gestures. That looks intentional, but it also means Safari/WebKit call sites stop being a special case when SimHID is available. I don't see a blocker in the diff itself, but I'd add an integration assertion (or a short doc note) proving that this precedence shift is acceptable for browser-facing flows too, not just native-app paths.

shaun0927:
Review pass (2026-04-16)

Verdict: REQUEST CHANGES

High-risk concern:
- `src/tools/native-input-backend.ts` now returns the SimHID backend unconditionally before the WebKit tier. That changes backend precedence globally, not just for native-app flows. In practice this can reroute connected Safari/WebView interactions away from `WebKitInputBackend` and into simulator-level HID injection without any Safari/WebView regression coverage in this PR.

I think this should be addressed before merge: either scope the re-enabled SimHID path to the contexts that are actually intended to use it, or add explicit Safari/WebView coverage that proves the new precedence is safe.

shaun0927:
Review pass:
- Verdict: REQUEST CHANGES
- HIGH: `src/tools/native-input-backend.ts` now returns SimHID before the WebKit tier, which changes backend priority for connected Safari sessions too, not just native-app flows. That is a meaningful routing change, and I don't see a Safari-specific regression test here proving we still prefer the intended backend for browser automation.
- Before merge, I'd like either (a) explicit Safari coverage for the new priority order, or (b) a guard that limits the SimHID fast-path to the contexts you actually want to flip.

shaun0927:

**Recommendation: COMMENT**

Main concern here is coverage rather than logic:
- The change re-enables the Tier-1 SimHID path based on unit tests, but the regression being referenced (#491) was a live simulator behavior problem. I’d be much more comfortable with at least one hosted/live assertion for tap/swipe on a booted simulator before treating the issue as fully closed.

shaun0927:
Review pass: the main thing I'd still like to see is a broader routing regression check.

Re-enabling the unconditional Tier-1 SimHID return path in `src/tools/native-input-backend.ts` changes precedence not just for native app call sites, but also for any path that passes a connected WebKit client. The updated unit tests cover the direct SimHID happy path, but they don't show that Safari/WebView call sites that intentionally depend on WebKit-backed input still behave as expected after this precedence flip.

shaun0927:

I don’t see an immediate blocking issue in the implementation, but I would still like one regression guard for the new precedence change in `getInputBackend()`: the unit test proves SimHID now wins even when a WebKit client is connected, but there isn’t an integration assertion covering a Safari/WebKit-connected page after that reorder.

If the intent is “SimHID always beats WebKit for coordinate gestures now,” I’d either add one live/integration check for that contract or spell it out explicitly in the PR description before merge.

shaun0927 (commented):
**Code Review: APPROVE** (P0=0, P1=0, P2=0)

Clean re-enablement of SimHID Tier-1 return path. Dead code (`simhid-gated`) properly removed, tests updated to verify positive-path behavior. Net -30 lines.

PR #569, opened 2026-04-16
feat(input): add HID_BRIDGE_MISSING error and swipe interpolation test

shaun0927:
## Summary

- Add `HID_BRIDGE_MISSING` error code to `InputBackendErrorCode` — when `tryCreateSimulatorKitHIDBackend()` cannot locate the sim-hid-bridge binary, it now throws `InputBackendError('HID_BRIDGE_MISSING')` with searched paths instead of silently returning `null`. Consistent with how `ax-bridge` handles the missing binary case (`BRIDGE_NOT_FOUND`). The routing layer already catches this and falls through to the next tier.
- Add unit test documenting that swipe delegates N-step interpolation to the Swift bridge (10 steps over duration) rather than decomposing in the Node wrapper.
- Fixes 2 unchecked verification items from #483 checklist: `헬퍼 바이너리 부재 시 InputBackendError (HIDBridgeMissing) 발생` and `swipe가 N단계 중간 좌표를 헬퍼에 전달`.

## Test plan

- [x] `npx jest --testPathPatterns=sim-hid-input-backend` — 32/32 passed
- [x] `npm run build` — green
- [ ] Verify routing layer catches `HID_BRIDGE_MISSING` and falls through (existing catch block at `native-input-backend.ts:892-896`)

### DIFF
--- a/src/tools/sim-hid-input-backend.ts
+++ b/src/tools/sim-hid-input-backend.ts
@@ -103,6 +103,7 @@ export type InputBackendErrorCode =
   | 'NOT_IMPLEMENTED'
   | 'SPAWN_TIMEOUT'
   | 'BRIDGE_NOT_FOUND'
+  | 'HID_BRIDGE_MISSING'
   | 'JSON_PARSE_FAILURE'
   | 'UNKNOWN';
 
@@ -356,5 +357,10 @@ export async function tryCreateSimulatorKitHIDBackend(): Promise<
       return new SimulatorKitHIDInputBackend(candidate);
     }
   }
-  return null;
+  const searched = candidates.map((c) => `  - ${c}`).join('\n');
+  throw new InputBackendError(
+    `sim-hid-bridge not found. Searched:\n${searched}\n` +
+      'Run npm run build or set OPENSAFARI_ALLOW_SWIFT_INTERPRETER=1 for dev mode.',
+    'HID_BRIDGE_MISSING',
+  );
 }
--- a/tests/unit/sim-hid-input-backend.test.ts
+++ b/tests/unit/sim-hid-input-backend.test.ts
@@ -129,6 +129,25 @@ describe('SimulatorKitHIDInputBackend', () => {
     expect(args).toEqual([DEVICE, 'swipe', '10', '20', '30', '40', '0.75']);
   });
 
+  test('swipe delegates N-step interpolation to bridge — single call with start/end coords', async () => {
+    // Design: The Node wrapper sends only start and end coordinates to the
+    // Swift bridge as a single `swipe` command. The bridge internally
+    // interpolates N intermediate points (default: 10 steps over the given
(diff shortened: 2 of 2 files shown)


shaun0927:

**Verdict: COMMENT**

One follow-up I'd recommend before merge: `tryCreateSimulatorKitHIDBackend()` now throws `HID_BRIDGE_MISSING` when no artifact is found, whereas the old contract was a soft `null` miss. The new error is clearer, but the `try*` name still reads like a non-throwing probe. If this contract change is intentional, I'd either rename the helper or add one more caller-audit regression test so we don't accidentally turn a missing bridge into a hard failure outside the currently-covered call sites.

shaun0927:
Review pass (2026-04-16)

Verdict: COMMENT

No blocking issue stood out in this diff. The new `HID_BRIDGE_MISSING` error is materially more actionable than a silent `null`, and the added unit coverage documents the single-call swipe contract well.

I would merge this as-is.

shaun0927:
Review pass:
- Verdict: COMMENT
- MEDIUM: `src/tools/sim-hid-input-backend.ts` now throws `HID_BRIDGE_MISSING` from `tryCreateSimulatorKitHIDBackend()` instead of returning `null` on the bridge-missing path. The current internal call sites look covered, but the `tryCreate*` name/signature still reads like a soft probe.
- Suggestion: either rename/document the function as a throwing probe, or add a short contract note near the export so future callers don't assume `null` is the only negative result.

shaun0927:

**Recommendation: COMMENT**

I didn’t find a blocking issue in the new `HID_BRIDGE_MISSING` path for the current in-repo callers.

Follow-up I’d still like after merge:
- Add one diagnose/tool-level assertion around this path as well. This helper used to signal absence via `null`, and now it is exception-based, so having one higher-level regression test would make the contract change safer over time.

shaun0927:
Review pass: one medium-risk contract issue to clean up before more call sites rely on it.

`src/tools/sim-hid-input-backend.ts` now throws `InputBackendError('HID_BRIDGE_MISSING')` when the helper is absent, but the exported signature still advertises `Promise<SimulatorKitHIDInputBackend | null>`. The routing layer and `diagnose` already catch the throw, so runtime behavior is OK, but the public contract is now misleading for any future caller that still assumes a nullable result.

shaun0927:

The explicit `HID_BRIDGE_MISSING` error makes the missing-binary path much more actionable than the previous silent `null` return.

Small follow-up only: `probeSimHid()` in `src/tools/diagnose.ts` still checks `if (!backend)` even though `tryCreateSimulatorKitHIDBackend()` now either returns a backend or throws. Not blocking, but cleaning that dead branch in the same rollout would keep the diagnostics contract aligned.

shaun0927 (commented):

P2: Return type of `tryCreateSimulatorKitHIDBackend` still includes `| null` but function now always throws or returns — stale type + dead branch in diagnose.ts. Not blocking (callers already use try/catch).

PR #575, opened 2026-04-16
feat(ci): cursor-monitor helper + wrap live smoke tests (#540)

shaun0927:
## Summary

- Adds `src/native/cursor-monitor.swift` — a small Swift helper that wraps a child process, polls `CGEventGetLocation` on a background thread, and exits non-zero if the host cursor drifts beyond a threshold (default 0.5px) during the child's lifetime.
- Adds `scripts/run-with-cursor-monitor.js` — a thin Node wrapper that prefers `dist/cursor-monitor` and falls back to `swift` interpreter.
- Wires the monitor into the three live jest invocations in `.github/workflows/headless-smoke.yml` (Flutter VM, Native simhid, WebView).

Refs: #540 — moves the acceptance-criteria box *"Integration tests never move the mouse (CI `CGEventGetLocation` monitor)"* from ☐ to demonstrable CI guard.

## Why

#540 requires integration tests to never reposition the host cursor. Today the workflow only enforces posture via env vars (`OPENSAFARI_HEADLESS_ONLY=1`, `OPENSAFARI_ALLOW_FOCUS_INPUT` unset). Those block construction-time AppleScript, but a future regression could still emit synthesized mouse events at runtime. A direct cursor-position assertion closes that loop.

## Design notes

- **Graceful fallback when no WindowServer** — if the initial `CGEventGetLocation` call returns NaN (headless VM without a virtual display), the monitor logs a warn line and runs the child unmonitored. This way the change is a safety-net, not a gate that can mis-fire.
- **Unconditional wrap vs opt-in env gate** — chose unconditional wrap so any accidental cursor-moving backend (e.g. someone re-adding an AppleScript shim) is caught from day one.
- **Binary lives in `dist/`** — builds alongside `ax-bridge` / `sim-hid-bridge` through the existing swiftc + codesign fallback pattern in `package.json` scripts.

## Test plan

- [x] `swiftc -O src/native/cursor-monitor.swift -o /tmp/cursor-monitor-test && /tmp/cursor-monitor-test -- /bin/echo hello` → reports `max_delta_px=0`, exits 0.
- [x] `node scripts/run-with-cursor-monitor.js -- /bin/echo wrapper` → interpreter-fallback path works, child stdout forwarded.
- [x] YAML validated: `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/headless-smoke.yml'))"`.
- [x] `npm run lint -- scripts/run-with-cursor-monitor.js` → 0 new errors.
- [ ] Scheduled cron run on `develop` after merge — expect `[cursor-monitor] ...` JSON line in each of the three live jest steps and `max_delta_px=0` throughout.


shaun0927 (commented):
Code review: P0=0, P1=0 (fixed in latest push), P2=1. Signal handler race fixed — handlers now registered before process.run(). 'goroutine' comment corrected. Clean to merge.

PR #576, opened 2026-04-16
refactor(ci): extract AppleScript audit-log assertion into a shared script (#540)

shaun0927:
## Summary

- New `scripts/assert-no-applescript.js` — reads `\$HOME/.opensafari/audit.log` (or `\$OPENSAFARI_AUDIT_LOG`) and exits non-zero if any line matches `applescript` case-insensitively; absent log is treated as pass (HEADLESS_ONLY=1 blocks construction time).
- `.github/workflows/headless-smoke.yml` — collapses three identical inline bash blocks (Flutter, Native simhid, WebView) into a single `node scripts/assert-no-applescript.js` call per job.
- New `tests/scripts/assert-no-applescript.test.ts` — 5 cases covering all script branches.

Refs: #540

## Why

All three \"Verify no AppleScript fallback was loaded\" steps carried the same 11-line bash. Any future evolution of the assertion (parse JSONL, assert \`backendKind\`, tolerate alternate audit paths) would need to stay in lockstep across three copies. Centralising it is a small DX improvement and unblocks the upcoming \`backendKind\` assertion that lands in a separate PR (unit C).

## Test plan

- [x] \`npx jest tests/scripts/assert-no-applescript.test.ts\` → 5/5 pass.
- [x] Manual: empty log → exit 0. Absent log → exit 0. Log with \`applescript\` → exit 1 with \`::error::\` annotation and offending lines printed.
- [x] \`python3 -c \"import yaml; yaml.safe_load(open('.github/workflows/headless-smoke.yml'))\"\` — YAML still valid after collapse.
- [ ] Scheduled cron run on \`develop\` after merge — three smoke jobs should still gate on AppleScript-free audit logs.

## Compatibility

Semantics match the bash implementation exactly:
- prints the same \`::error::\` annotation string
- prints at most 20 offending lines (previously \`grep -n ... | head -20\`)
- absent log is a pass (identical bash behaviour)

### DIFF
--- a/.github/workflows/headless-smoke.yml
+++ b/.github/workflows/headless-smoke.yml
@@ -401,19 +401,7 @@ jobs:
 
       - name: Verify no AppleScript fallback was loaded
         if: always()
-        run: |
-          set -euo pipefail
-          LOG="$HOME/.opensafari/audit.log"
-          if [ ! -f "$LOG" ]; then
-            echo "audit log not present — HEADLESS_ONLY=1 blocks the AppleScript backend at construction time."
-            exit 0
-          fi
-          if grep -q -i 'applescript' "$LOG"; then
-            echo "::error::audit.log references applescript — headless posture violated."
-            grep -n -i 'applescript' "$LOG" | head -20
-            exit 1
-          fi
-          echo "audit log free of applescript references"
+        run: node scripts/assert-no-applescript.js
 
       - name: Capture screenshot on failure
         if: failure()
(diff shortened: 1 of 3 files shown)


shaun0927 (commented):
Code review: P0=0, P1=0, P2=1 (cosmetic ::error:: filename). Clean refactor with exact semantic equivalence. 5/5 tests cover all branches. LGTM.

PR #577, opened 2026-04-16
test(ci): audit.log posture helper + jest suites for backendKind allow-list (#540)

shaun0927:
## Summary

- New `src/ci/audit-log-posture.ts` — tolerant parser that pulls every `backendKind` value out of `~/.opensafari/audit.log` (or `\$OPENSAFARI_AUDIT_LOG`) whether it lives at the top level or inside a stringified `args_summary`, and exposes `scanAuditLog()` / `assertAuditLogPosture()`.
- New `tests/unit/audit-log-posture.test.ts` — 10 fixture-driven cases covering all branches (10/10 pass locally).
- New `tests/integration/audit-log-backend-kind.live.test.ts` — runs against the real audit log produced by the live smoke jobs; default `npm test` skips it (excluded in `jest.config.js`).

Refs: #540 — moves the acceptance-criteria box *"Telemetry: `backendKind in {'flutter-vm','simhid','webkit'}` only (`applescript` = 0)"* into a shape that can be asserted in CI. Workflow wiring is a follow-up so this PR stays a pure library addition.

## Why

Unit B (PR #576) already centralises the \"no applescript\" check. This PR adds the positive-side constraint — every recorded backend must come from the allowed set — with a testable helper that the smoke jobs can call after their live steps land.

## Test plan

- [x] \`npx jest tests/unit/audit-log-posture.test.ts --no-coverage\` → 10/10 pass.
- [x] Parser handles \`\"backendKind\":\"simhid\"\` and \`\\\"backendKind\\\":\\\"simhid\\\"\` shapes.
- [x] \`scanAuditLog('/does/not/exist')\` returns a zero-entry report (treats absent log as vacuous pass).
- [ ] After merge, follow-up PR wires the live suite into each smoke job so the assertion runs on every scheduled cron.

## Known red CI

\`develop\` currently fails \`npm run build\` because of a duplicate \`wsToHttpUrl\` in \`src/flutter/vm-service-discovery.ts\` (introduced by commit \`5527d3f7\`). That is unrelated to this PR but will break this PR's CI until a hotfix lands. I'm opening that hotfix as a separate PR immediately after.

### DIFF
--- a/src/ci/audit-log-posture.ts
+++ b/src/ci/audit-log-posture.ts
@@ -1,0 +1,118 @@
+/**
+ * Helpers for asserting on the posture of `~/.opensafari/audit.log`.
+ *
+ * The audit log is JSONL. We intentionally keep the scan tolerant of the
+ * exact schema so regressions that add new fields (or move `backendKind`
+ * between top-level and nested positions) do not silently pass.
+ *
+ * Two signals matter for the #540 acceptance criteria:
+ *
+ *   1. `applescriptHits` — any entry that mentions `applescript` (case
+ *      insensitive) is a headless-posture violation when
+ *      `OPENSAFARI_HEADLESS_ONLY=1` is set.
+ *   2. `backendKinds` — the set of backend identifiers we can extract from
+ *      the log, regardless of whether they live at the top level or inside
+ *      a stringified `args_summary`. The allowed set is
+ *      `{'flutter-vm','simhid','webkit'}`.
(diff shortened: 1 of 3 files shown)


shaun0927 (commented):
Code review: P0=0, P1=0, P2=2 (vacuous pass semantics + path traversal — both CI-only). Well-structured posture helper with 10 thorough unit tests. LGTM.

Issue #574, opened 2026-04-16
ci: cross-Xcode / macOS / device / Flutter matrix for headless-smoke


shaun0927:
## Context

Spun out of epic #540 (successor to #484). The Matrix section of the epic requires green runs on:

- macOS 14, 15, 16
- Xcode 15, 16, 17, 26
- iPhone 15 / 16 / 17 / Pro / iPad
- Flutter 3.x, 4.x

Today `headless-smoke.yml` runs on `macos-latest` only with whichever Xcode that runner image ships.

## Problem

The current single-dimension smoke cannot surface breakage that is specific to an older or newer combination. The epic #540 closure rule (30 consecutive green days) cannot start counting on the full matrix until the matrix actually exists.

## Proposed Scope

1. Introduce a matrix strategy in `headless-smoke.yml` for Safari / Flutter / Native / WebView smoke jobs.
2. Start with one baseline combination per dimension:
   - macOS 15 (current `macos-latest`), Xcode 17, iPhone 16, Flutter 3.x.
3. Add each additional cell behind `workflow_dispatch` input flags so we can stage the rollout without exploding cron cost.
4. Flip cells on as runner images become available and the job proves stable.
5. Track cost: record the per-job minutes in the workflow summary so we can budget ahead of the 30-day lock-in.

## Files

| File | Change |
|---|---|
| `.github/workflows/headless-smoke.yml` | Add `strategy.matrix` + `workflow_dispatch.inputs` gating |
| `docs/ci/headless-smoke-matrix.md` | **NEW** — rollout plan and runner availability notes |

## Constraints

- No new required cells must be added before the runner image is actually available — otherwise the cron breaks red and starts the 30-day countdown from zero.
- Must not require `OPENSAFARI_ALLOW_FOCUS_INPUT` on any cell.
- Must not require Simulator.app foreground on any cell.

## Exit criteria

- Every cell in the epic #540 Matrix section has at least one green run.
- Matrix is the default configuration (no `if: false` gates left on the required cells).
- Epic #540 "Cross-Xcode matrix" box flips to ✅.

Refs: #540, #484

shaun0927:
Closing as out-of-scope for now.

[...]

PR #579, opened 2026-04-16
ci: add matrix skeleton to headless-smoke jobs (#540)

shaun0927:
## Summary

Adds a single-cell \`strategy.matrix\` stanza to all four smoke jobs (safari / flutter / native / webview) in \`.github/workflows/headless-smoke.yml\`.

- \`matrix.os: [macos-latest]\` — baseline cell, behaviour unchanged.
- \`fail-fast: false\` — future sibling cells won't kill each other once the matrix grows.
- \`runs-on: \${{ matrix.os }}\` on all four jobs.

Refs: #540 (acceptance-criteria Matrix section) + #574 (cross-Xcode matrix child issue).

## Why

Epic #540's Matrix section requires green runs across macOS 14/15/16, Xcode 15/16/17/26, iPhone 15/16/17/Pro/iPad, and Flutter 3.x/4.x. Each dimension will be rolled out as runner images become available. Landing the skeleton now means the rollout is a one-line append to \`matrix.os\` (or a new matrix dimension) rather than a full job duplication.

## Why one cell now, more later

Adding additional cells (macos-14, macos-15, macos-16) before the runner images are proven stable would keep the cron red and reset the 30-day green countdown in the epic's closing rule. Child issue #574 tracks the staged rollout.

## Test plan

- [x] \`python3 -c \"import yaml; yaml.safe_load(open('.github/workflows/headless-smoke.yml'))\"\` — YAML valid.
- [x] \`git diff\` touches only the four job headers; no step logic moved.
- [x] Job \`name:\` fields kept unchanged so any existing branch-protection status-check names still match once the single-cell matrix kicks in (GitHub auto-appends \`(macos-latest)\` only when multiple cells exist).
- [ ] Cron run on \`develop\` after merge — same green signal as today's single-runner job.

### DIFF
--- a/.github/workflows/headless-smoke.yml
+++ b/.github/workflows/headless-smoke.yml
@@ -27,7 +27,16 @@ env:
 jobs:
   safari-smoke:
     name: Safari — headless QA audit
-    runs-on: macos-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        # Baseline cell. Additional cells (macos-14, macos-15, macos-16,
+        # Xcode / device / Flutter-version dimensions) land in follow-up PRs
+        # — adding them before the runner images are proven stable would
+        # keep the cron red and reset the 30-day green countdown from #540.
+        # See epic #540 + child issue #574 for the staged rollout plan.
+        os: [macos-latest]
     timeout-minutes: 20
     steps:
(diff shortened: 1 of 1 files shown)


shaun0927 (commented):
Code review: P0=0, P1=0, P2=0. Trivial matrix scaffolding. Functionally identical to prior runs-on. LGTM.

PR #580, opened 2026-04-16
fix(ci): surface ax-bridge codesign and exec failures (#501)

shaun0927:
## Summary

Diagnostics-only PR for the headless smoke cascade failure on GitHub-hosted runners. The WebView job's 7 tests and Flutter job's 4 tests all fail at the same Swift bridge invocation point, but the existing build pipeline and TS wrapper hide *why*:

- \`npm run build:ax-bridge\` redirects \`swiftc\` and \`codesign\` stderr to \`/dev/null\` and **deletes the unsigned binary on codesign failure**, forcing every run through a \`swift\` interpreter fallback whose own failure surface is just as opaque.
- \`AccessibilityBridge.exec()\` throws \`Command failed: <cmd>\` without forwarding the bridge's stderr, so each failure mode (unsigned binary, missing TCC, simulator not booted, swift compile error) reads identically in the jest output.

## Changes

| File | Change |
|---|---|
| \`package.json\` | Drop \`2>/dev/null\` from \`build:ax-bridge\` and \`build:sim-hid-bridge\`; keep the unsigned binary on codesign failure (ad-hoc signing is hardening, not a runtime requirement for AX queries against locally-booted sims). |
| \`src/native/accessibility-bridge.ts\` | Append the last 5 stderr lines and the failing command to \`AccessibilityBridgeError\`, fix a stale catch that swallowed re-thrown bridge errors. |

## Why this is small

This PR doesn't try to make WebView or Flutter pass yet — that requires either fixing whatever the underlying TCC/permission issue actually is, or moving to a self-hosted runner with the right entitlements. **The next CI run on this branch will name the root cause in the failure message**, which is the prerequisite for any further fix landing meaningfully.

## Test plan

- [ ] \`Headless Smoke Tests\` workflow on this PR shows the *real* swiftc/codesign error in the build step (or confirms the unsigned binary is now retained and runs).
- [ ] WebView and Flutter jest output now includes \`stderr: ...\` tail in the \`AccessibilityBridgeError\` message.
- [ ] \`npm test\` (unit suite) stays green.

Towards: #501


shaun0927 (commented):
Code review: P0=0, P1=0, P2=0. Fixes a real catch-block bug in accessibility-bridge.ts. Removes stderr suppression from build. LGTM.

PR #526, opened 2026-04-15
ci: add Flutter headless smoke job (#501, part 2/4)

shaun0927:
## Summary

- Adds `flutter-smoke` job to `.github/workflows/headless-smoke.yml`.
- Installs Flutter via `subosito/flutter-action`, boots an iPhone simulator, launches `tests/integration/fixtures/flutter_sample` via `flutter run --machine` (DDS required per v0.4.7 notes), and runs `tests/integration/flutter-vm-input.live.test.ts` with `OPENSAFARI_LIVE_VM=1`.
- Exercises the Tier-0 `FlutterVMInputBackend` end-to-end. The live test deliberately does NOT set `OPENSAFARI_ALLOW_FOCUS_INPUT`; combined with `OPENSAFARI_HEADLESS_ONLY=1` from PR 1's workflow env, any silent fallback to AppleScript/CGEvent would throw `HeadlessInputUnavailableError` and fail the job.
- Emits JUnit via a new `jest-junit@^16` devDep; uploads JUnit + `flutter-run.log` on every run; captures simulator screenshot on failure.
- Extra belt-and-braces check: scans `~/.opensafari/audit.log` for any `applescript` reference and fails the job if one appears.

## Stacked PR

> ⚠️ **This PR is stacked on [#524](https://github.com/shaun0927/opensafari/pull/524) (part 1/4).** Base branch is `feat/501-ci-headless-smoke-safari`. Once #524 merges, please rebase/retarget this PR to `develop`.

## PR sequence for #501

| # | Scope | Status |
|---|---|---|
| 1 | Workflow skeleton + Safari job | #524 |
| 2 | Flutter smoke job + jest-junit | **this PR** |
| 3 | Native smoke job (Settings.app) | follow-up |
| 4 | WebView smoke job (or separate fixture issue) | follow-up |

## Verification Checklist mapping (issue #501)

Adds coverage for:

- [x] Flutter 자동화 — 샘플 앱 멀티스텝 시나리오 통과
- [x] `OPENSAFARI_ALLOW_FOCUS_INPUT` 미설정 상태 (asserted pre-flight)
- [x] Simulator.app foreground 강제 없음 (workflow never launches Simulator.app)
- [x] telemetry에서 `backendKind=flutter-vm`만 사용 (`applescript` 0회 — audit-log grep guard)
- [x] JUnit 리포트 출력
- [x] 실패 시 스크린샷 아티팩트 업로드
[...]

PR #581, opened 2026-04-16
fix(ci): enable Flutter DDS frontend compiler on CI smoke (#501)

shaun0927:
## Summary

The Flutter smoke job has been failing all 4 jest tests on every CI run since #526 landed, hidden by \`continue-on-error: true\`. This PR fixes the root cause and surfaces real failures.

### Root cause

From the most recent CI failure log (run \`24506966387\`):

\`\`\`
[input-backend] Flutter VM on <UDID> rejects evaluate (code 113).
Likely a release build or \`simctl launch\` without \`flutter run\` —
falling back past Tier 0.
\`\`\`

\`flutter run --no-hot --machine\` was the workflow's compromise to keep the launcher non-blocking without stdin, but **\`--no-hot\` disables the DDS frontend compiler service** that \`FlutterVMInputBackend.probeEvaluateCompile\` calls. Without it the backend rejects the device, \`getInputBackend()\` falls through to \`HeadlessInputUnavailableError\`, and every Tier-0 input test errors out.

The previous comment on the workflow noted *\"hot-reload mode hangs the DDS connection on CI\"* — but the actual symptom of plain \`flutter run\` was that the launcher blocked on stdin, not that DDS itself hung. Fixing that with \`</dev/null\` lets us keep the frontend compiler enabled.

### Changes

| Aspect | Before | After |
|---|---|---|
| flutter run flags | \`--no-hot --machine\` | \`--machine --vmservice-out-file <path>\` (stdin → /dev/null) |
| VM service URL discovery | regex scrape on \`flutter --machine\` JSON stream (4 fallback patterns) | read \`--vmservice-out-file\` (written only after engine binds the port) |
| DDS warmup | warning on failure (\`grep -q \"Dart\"\`) | hard error on failure (\`HTTP 200\` check + 8s frontend-compiler settle) |
| Failure visibility | \`continue-on-error: true\` | removed — failures surface in the workflow result |

### Why this should now actually pass

\`probeEvaluateCompile\` invokes \`compileExpression\` over JSON-RPC, which the DDS frontend compiler service answers. With hot-reload mode enabled the service is up by the time the warmup probe sees HTTP 200, and the 8-second settle covers the initial compile pass. The other three input tests (\`tap\` / \`typeText\` / \`swipe\`) only need \`getInputBackend()\` to return Tier-0 — once \`probeEvaluateCompile\` succeeds, that path is unblocked.

## Test plan

- [ ] Headless Smoke Tests workflow on this PR shows \`flutter-smoke\` job genuinely green (not \`continue-on-error\`-masked).
- [ ] Job log shows \`VM service file written after Ns\` and \`DDS HTTP endpoint responsive\` from the new probe.
- [ ] Jest output shows 4 passed, 0 failed for \`flutter-vm-input.live.test.ts\`.

Towards: #501

### DIFF
--- a/.github/workflows/headless-smoke.yml
+++ b/.github/workflows/headless-smoke.yml
@@ -292,9 +292,23 @@ jobs:
         run: |
           set -euo pipefail
           mkdir -p "$GITHUB_WORKSPACE/flutter-smoke-logs"
+          # Use --vmservice-out-file so we get the resolved URL from a file
+          # instead of scraping the --machine JSON stream. The previous
+          # approach combined `--no-hot` (to keep `flutter run` non-blocking
+          # without stdin) with log-scraping for the VM service URL, but
+          # `--no-hot` disables the DDS frontend compiler service that
+          # FlutterVMInputBackend's `probeEvaluateCompile` requires
+          # (returns error 113 → backend is rejected → no Tier-0 routing).
+          # `--vmservice-out-file` works in both hot and no-hot modes and
+          # avoids the parse fragility entirely.
+          VM_SERVICE_FILE="$GITHUB_WORKSPACE/flutter-smoke-logs/vmservice.uri"
+          rm -f "$VM_SERVICE_FILE"
           pushd "$FIXTURE_DIR" >/dev/null
           flutter pub get
-          nohup flutter run -d "$SIMULATOR_UDID" --no-hot --machine \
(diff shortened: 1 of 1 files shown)


shaun0927 (commented):
Code review: P0=0, P1=0, P2=1 (timing). Root-cause fix for Flutter DDS compiler. --vmservice-out-file replaces fragile log scraping. Removes continue-on-error. LGTM.

PR #578, opened 2026-04-16
fix(build): remove duplicate wsToHttpUrl function (#540)

shaun0927:
## Summary

\`develop\` is currently failing CI on every subsequent push/PR because commit \`5527d3f7\` accidentally re-declared \`wsToHttpUrl\` in \`src/flutter/vm-service-discovery.ts\`. ts-loader surfaces this as TS2323 + TS2393 during \`npm run build\`, which fails the \`lint\` and \`test\` jobs (both run \`npm run build\` through the \`prepare\` hook — see \`.github/workflows/ci.yml\`).

This PR removes the undocumented duplicate (lines 89-92) and keeps the documented one (lines 99-102).

## Evidence

Before:
\`\`\`
src/flutter/vm-service-discovery.ts(89,17): error TS2323: Cannot redeclare exported variable 'wsToHttpUrl'.
src/flutter/vm-service-discovery.ts(89,17): error TS2393: Duplicate function implementation.
server (webpack 5.105.4) compiled with 5 errors in 3458 ms
\`\`\`

After:
\`\`\`
server (webpack 5.105.4) compiled successfully in 3391 ms
\`\`\`

## Test plan

- [x] \`npm run build:src\` — 0 errors.
- [x] \`npx jest tests/unit/flutter-vm-service.test.ts --no-coverage\` — 21/21 pass.
- [x] \`git grep wsToHttpUrl\` confirms no caller relied on the duplicate's definition (both were byte-identical).

## Why this blocks #540 work

Three in-flight PRs (#575 / #576 / #577) inherit this broken state — their CI will stay red until this lands. Those PRs will rebase after merge.

Refs: #540

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #582, opened 2026-04-16
perf(metrics): single-syscall memory sampling on timedInput hot path (#554)

shaun0927:
## Summary

- Deduplicate `process.memoryUsage()` syscalls on the per-op hot path so the `timedInput` memory sampler stays under the 50 µs budget called out in #554.
- Add `recordMemorySampleFromRss(rssBytes)` in `memory-tracker` for callers that already have a fresh RSS reading (peak / time-series / soft-cap watchdog all reuse a single sample).
- Update `emitInputTelemetry` to take the new path whenever the event already carries `rss_mb`; events without memory fields still call `recordMemorySample()` so non-timedInput paths keep ticking.
- Rewrite the microbench to measure the optimized single-syscall sequence and assert directly against the `< 50 µs` budget.

## Why

The pre-existing microbench exercised two `process.memoryUsage()` / `rss()` syscalls per iteration (one in `sampleMemoryFields`, another in the follow-up `recordMemorySample`). On a loaded macOS host the per-call overhead ran 68–157 µs across 3 independent runs — consistently above the issue #554 budget. Reusing a single sample removes the second syscall and lands us well under budget.

## Checklist item closed

- \`[x]\` **Instrumentation → No measurable latency regression (< 50 µs overhead per call, measured by a microbench)** (#554)

## Test plan

- [x] \`npx jest tests/unit/memory-tracker.test.ts tests/unit/input-telemetry.test.ts tests/unit/input-telemetry-rollup.test.ts tests/unit/diagnose.test.ts\` — 67/67 green (microbench included, previously 1 failing)
- [ ] \`npm run build\` green on CI
- [ ] Peak-tracker behaviour unchanged when \`OPENSAFARI_INPUT_TELEMETRY_MEMORY=0\` is set (new \`recordMemorySampleFromRss\` early-returns via the same kill-switch)

### DIFF
--- a/src/metrics/input-telemetry.ts
+++ b/src/metrics/input-telemetry.ts
@@ -14,7 +14,12 @@
 import { AsyncLocalStorage } from 'node:async_hooks';
 import type { InputBackendKind } from '../tools/native-input-backend';
 import { accumulateInputTelemetry } from './input-telemetry-rollup';
-import { recordMemorySample, bytesToMB, isMemoryTrackingEnabled } from './memory-tracker';
+import {
+  recordMemorySample,
+  recordMemorySampleFromRss,
+  bytesToMB,
+  isMemoryTrackingEnabled,
+} from './memory-tracker';
 
 /**
  * Stable set of operations we time. Matches the `InputBackend` interface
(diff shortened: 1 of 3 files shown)


shaun0927 (commented):
CI all green (lint/test/build). Rebased cleanly onto develop (after #578 merge). Microbench passes at ~20-30 µs vs 50 µs budget via single-syscall optimization.

PR #583, opened 2026-04-16
feat(ci): add explicit Safari tap/type/scroll smoke scenarios (#501)

shaun0927:
## Summary

Epic #484's acceptance criterion for the Safari smoke is *\"Safari 자동화 — 임의 페이지 탭/입력/스크롤 통과\"*, but the existing \`safari-smoke\` job only runs \`opensafari audit\` (a one-shot QA scan). No explicit tap/type/scroll interaction was ever exercised, so the corresponding checkbox on #501 could not be flipped honestly even though the job was green.

## Changes

| File | Change |
|---|---|
| \`tests/integration/safari-smoke.live.test.ts\` (NEW) | Live integration suite with tap (button click), type (input event dispatch), and scroll (\`window.scrollTo\` + \`scrollY\` assertion). Fixture HTML is injected into https://example.com via \`document.body.innerHTML\` so the assertions don't depend on whatever the upstream page ships today. |
| \`.github/workflows/headless-smoke.yml\` | New \`Run Safari explicit tap/type/scroll scenarios\` step in the \`safari-smoke\` job, runs after the audit. JUnit emitted into the existing \`safari-smoke-junit\` artifact. |

## Why this works headlessly

All three interactions route through the WebKit Remote Debugging Protocol's \`Runtime.evaluate\`, which is the \`webkit\` backend kind by definition — no AX bridge invocation, no SimulatorKit HID, no AppleScript. This satisfies the existing \`Verify no AppleScript fallback was loaded\` step as a side-effect, and is the same protocol the WebView smoke job uses for its DOM assertions.

## Test plan

- [ ] \`Headless Smoke Tests\` workflow on this PR runs the new step and reports 4 passed (tap, type, scroll, routing).
- [ ] \`safari-smoke-junit\` artifact contains both \`safari-audit.junit.xml\` and \`safari-scenarios.junit.xml\`.
- [ ] Scenario test does NOT use ax-bridge / sim-hid-bridge / AppleScript (verified by the existing audit-log probe).

Towards: #501

### DIFF
--- a/.github/workflows/headless-smoke.yml
+++ b/.github/workflows/headless-smoke.yml
@@ -181,6 +181,30 @@ jobs:
             --format junit \
             --output qa-results/safari-audit.junit.xml
 
+      - name: Run Safari explicit tap/type/scroll scenarios
+        env:
+          OPENSAFARI_LIVE_VM: '1'
+          OSF_DEVICE_ID: ${{ env.SIMULATOR_UDID }}
+          SAFARI_SMOKE_URL: https://example.com
+          JEST_JUNIT_OUTPUT_DIR: qa-results
+          JEST_JUNIT_OUTPUT_NAME: safari-scenarios.junit.xml
+          JEST_JUNIT_CLASSNAME: '{filepath}'
+          JEST_JUNIT_SUITE_NAME: 'Safari headless smoke (tap/type/scroll)'
+        run: |
+          set -euo pipefail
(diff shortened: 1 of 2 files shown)


shaun0927 (commented):
Code review: P0=0, P1=0, P2=2 (low). Clean Safari integration test with proper opt-in guard, fixture injection, and routing assertion. LGTM.

Issue #573, opened 2026-04-16
arch: AX bridge GUI-less path — decouple AX reads from Simulator.app foreground


shaun0927:
## Context

Spun out of epic #540 (successor to #484). The current `src/native/ax-bridge` pipeline reads the Accessibility tree via AX APIs that are served by Simulator.app. Even after the input side moves to SimulatorKit HID / (future) SimDeviceIO, the **read** side still goes through the AX bridge, which silently assumes Simulator.app is running in front.

## Problem

#540 "No job requires Simulator.app foreground" cannot be proven until AX reads also work without Simulator.app at the front. Ideally the AX channel should use one of:

- `SimDevice.io` stream with an AX replay target, or
- a private `SimulatorKit` AX endpoint exposed to the host process.

## Proposed Scope

1. Survey which private AX endpoints `SimulatorKit` or `CoreSimulator` expose to host processes for a running SimDevice.
2. Build a PoC reader that returns the same shape as `src/native/ax-bridge.swift` without requiring Simulator.app to be foreground.
3. Gate the new path behind a runtime probe so we fall back to the existing AX bridge when the private endpoints are unavailable.
4. Extend the private-API sentinel (`.github/workflows/private-api-sentinel.yml`) to watch the symbols we rely on.

## Files (tentative)

| File | Change |
|---|---|
| `docs/adr/gui-less-ax-bridge.md` | **NEW** — ADR |
| `src/native/ax-bridge-gui-less.swift` | **NEW** — PoC |
| `src/tools/ax-reader.ts` | Extend to prefer GUI-less endpoint when probe succeeds |
| `.github/workflows/private-api-sentinel.yml` | Add symbols |

## Constraints

- Behaviour-preserving: the tree shape and field names must match the current AX bridge output.
- No regression on Xcode 15/16/17/26 when the new endpoint is absent.

## Exit criteria

- ADR with a "go" / "no-go" decision.
- If "go": live test proves AX reads succeed with Simulator.app **not** foreground.
- Epic #540 still progresses toward "No job requires Simulator.app foreground".

Refs: #540, #484
Blocks: #540 ("AX bridge GUI-less path" box)

shaun0927:
Closing as completed — all exit criteria met by PR #587 (merged 2026-04-17).

## Verification (live, via opensafari MCP)

- macOS frontmost app during verification: **KakaoTalk** (not Simulator.app).
[...]

PR #587, opened 2026-04-17
docs(573): ADR — existing ax-bridge is already GUI-less (no-go on new PoC)

shaun0927:
## Summary

- Records a "no-go" ADR on the separate `src/native/ax-bridge-gui-less.swift` PoC that issue #573 originally scoped.
- The existing `src/native/ax-bridge.swift` resolves the AX root with `AXUIElementCreateApplication(pid)`, which targets Simulator.app by PID and does **not** require it to be the macOS frontmost app.
- Documents the surveyed alternatives (private `SimDevice.io` AX replay, `SimulatorKit` AX endpoint, `AXUIElementCreateSystemWide`) and why each was rejected.

## Empirical evidence

Reproduced on `iPhone 17 Pro` / iOS 26.4, Ghostty forced `frontmost=true`:

```
$ /dist/ax-bridge dump --device <UDID> --max-depth 2
→ exit 0 · full tree including iOS content group (Safari back/reload/address)
```

The MCP-host failure observed at first glance was a per-bundle TCC grant issue on the Claude Code host, not a bridge design issue (see §2.2 of the ADR).

## Follow-ups

- A separate PR adds sentinel probe 7 that enforces the GUI-less property on every daily run, so the implicit guarantee cannot silently regress.

## Test plan

- [x] ADR renders on GitHub without broken internal links.
- [x] No source or workflow files touched — pure docs-only change.
- [ ] Reviewer confirms "no-go" rationale covers the #573 exit criteria table (§6).

Closes part of #573 · Refs #540 · Predecessor #484

### DIFF
--- a/docs/adr/gui-less-ax-bridge.md
+++ b/docs/adr/gui-less-ax-bridge.md
@@ -1,0 +1,181 @@
+# ADR: GUI-less AX Bridge Path
+
+> Status: **Accepted — "no-go" on separate PoC.** The existing
+> `src/native/ax-bridge.swift` is already GUI-less: it targets the
+> Simulator.app process by PID via `AXUIElementCreateApplication(pid)`, which
+> does not require Simulator.app to be the macOS frontmost application.
+>
+> Tracks: [#573](https://github.com/shaun0927/opensafari/issues/573) ·
+> Epic: [#540](https://github.com/shaun0927/opensafari/issues/540) ·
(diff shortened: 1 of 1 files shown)


shaun0927 (commented):
Self-review: no-go ADR is the correct call — the existing AX bridge already satisfies the GUI-less read invariant via AXUIElementCreateApplication(pid), and a new private-API surface would enlarge the sentinel scope without capability gain. Regression protection moves to sentinel probe 7 (#616). LGTM.

PR #612, opened 2026-04-17
fix(540): admit ax-press to audit-log backendKind allow-list

shaun0927:
## Summary

- Adds `ax-press` to `ALLOWED_BACKEND_KINDS` in `src/ci/audit-log-posture.ts` alongside `flutter-vm`, `simhid`, and `webkit`.
- Leaves `applescript` and `simctl` rejected — the allow-list is still narrow, just now correct for the backends that actually exist.
- Updates the unit test that previously asserted `ax-press` was disallowed (it now has a positive test instead) and refreshes the doc-comment in the live posture test.

## Why

The 2026-04-17 verification pass on epic #540 noted:

> the newly-added `ax-press` backend is not in the AC whitelist `{flutter-vm, simhid, webkit}` and will need the AC updated or the backend excluded.

PR #587's ADR on issue #573 concluded the ax-bridge AX channel is already GUI-less — `AXUIElementCreateApplication(pid)` targets Simulator.app by PID and does not require it to be the macOS frontmost app. `src/tools/app-tap-element.ts` stamps `headless: true` on every `ax-press` result, so treating it as a first-class headless backend matches the observed posture. "AC updated" is the right side of that fork.

## Test plan

- [x] `npx jest --config jest.config.js tests/unit/audit-log-posture.test.ts` — 11 / 11 passing
- [x] Full jest unit suite — green locally
- [x] `npx eslint` clean on the three touched files

Refs: #540 · Follows: #587 (#573 ADR)

### DIFF
--- a/src/ci/audit-log-posture.ts
+++ b/src/ci/audit-log-posture.ts
@@ -20,10 +20,18 @@ import * as fs from 'node:fs';
 
 /**
  * The only `InputBackendKind` values that satisfy the #540 acceptance
- * criterion. Keep this list narrow — any legacy backend (`applescript`,
- * `simctl`, `ax-press`) should be rejected by the posture check.
+ * criterion. Keep this list narrow — legacy foreground-coupled backends
+ * (`applescript`, `simctl`) must stay rejected.
+ *
+ * `ax-press` is allowed because PR #587's ADR on #573 established that the
+ * ax-bridge AX channel is already GUI-less: `AXUIElementCreateApplication(pid)`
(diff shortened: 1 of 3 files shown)


shaun0927 (commented):
Self-review: admitting ax-press to the audit-log allow-list is consistent with PR #587's ADR (ax-bridge is already GUI-less) and the telemetry path in app_tap_element. Tests cover both the positive (ax-press accepted) and negative (applescript still rejected) cases. LGTM.

PR #613, opened 2026-04-17
feat(540): npm run test:live — DX-facing integration target with cursor-monitor guard

shaun0927:
## Summary

- Adds `npm run test:live` to `package.json` — sets `OSF_LIVE=1` and wraps the full integration suite with `scripts/run-with-cursor-monitor.js`.
- Documents the new DX target in `tests/integration/README.md`, explaining that the wrapper pipes the child through the `cursor-monitor` Swift helper which polls `CGEventGetLocation` and fails the run the moment the macOS cursor moves.
- Keeps `test:integration:monitored` as the non-gated alias for ad-hoc use without the `OSF_LIVE` signal.

## Why

Epic #540 acceptance criteria call for:

- `OSF_LIVE=1 npm run test:integration` green on local dev
- Integration tests never move the mouse (CI `CGEventGetLocation` monitor)

The wrapper and the Swift cursor-monitor already exist, but no top-level npm script surfaced them as the canonical developer command. This patch is the small naming gap between the existing infra and the AC text — no behavioural change to the underlying suite, just a reliable DX entry point.

## Test plan

- [x] `npm run test:live` prints the expected `OSF_LIVE=1` env line and invokes the cursor-monitor wrapper before delegating to `jest`.
- [x] `test:integration:monitored` still resolves and runs unchanged.

Refs: #540

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "test:unit": "jest --config jest.config.js",
     "test:integration": "jest --preset ts-jest --testEnvironment node --roots tests/integration --testPathIgnorePatterns '/node_modules/' --testPathIgnorePatterns 'webkit-connection' --testTimeout=120000 --runInBand --reporters=default --reporters=<rootDir>/tests/integration/screenshot-failure-reporter.cjs",
     "test:integration:monitored": "node scripts/run-with-cursor-monitor.js -- npm run test:integration",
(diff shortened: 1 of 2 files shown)


shaun0927 (commented):
Self-review: npm run test:live is the clean DX target for #540 — OSF_LIVE=1 + cursor-monitor wrapper gives local devs the same headless-posture guarantee CI enforces. Single npm script + README entry, zero src impact. LGTM.

PR #616, opened 2026-04-17
test(573): sentinel probe 7 enforces the ax-bridge GUI-less invariant

shaun0927:
## Summary

Adds the seventh private-API sentinel probe to prevent the ax-bridge GUI-less invariant from silently regressing. The invariant — that AX reads go through `AXUIElementCreateApplication(pid)` and therefore do not require Simulator.app to be the macOS frontmost app — was recorded in PR #587's ADR on issue #573.

- Static check: the Swift source at `src/native/ax-bridge.swift` must call `AXUIElementCreateApplication(` and must not call `AXUIElementCreateSystemWide(`. Always runs, regardless of host state.
- Runtime check: when a simulator is booted and Simulator.app is not the frontmost macOS app, `ax-bridge dump` must return non-empty JSON. Skips gracefully (with a warning) when no simulator is booted, when Simulator.app happens to be frontmost locally, or when the bridge returns `SIMULATOR_NOT_RUNNING`.

Workflow doc block in `.github/workflows/private-api-sentinel.yml` is extended from "six probes" to "seven probes" so the daily sentinel run reports the new coverage.

## Why

The PR #587 body promises: *"A separate PR adds sentinel probe 7 that enforces the GUI-less property on every daily run, so the implicit guarantee cannot silently regress."* This is that PR.

Without this probe, a future refactor of `ax-bridge.swift` that switched to `AXUIElementCreateSystemWide()` (or otherwise keyed AX reads off the frontmost app) would pass all existing tests, pass probe 4, and break epic #540's "No job requires Simulator.app foreground" acceptance criterion.

## Test plan

- [x] `npx jest --config jest.sentinel.config.js -t "ax-bridge GUI-less"` — probe 7 static assertions pass; runtime probe skips cleanly on hosts without a booted simulator.
- [x] `npx eslint tests/sentinel/private-api-probe.test.ts` — clean.
- [ ] Scheduled `private-api-sentinel.yml` run — will be exercised on the next cron or `workflow_dispatch`.

Refs: #573, #540 · Follows: #587

### DIFF
--- a/.github/workflows/private-api-sentinel.yml
+++ b/.github/workflows/private-api-sentinel.yml
@@ -12,6 +12,9 @@ name: Private API Sentinel
 #   4. ax-bridge AXUIElement dump returns non-empty JSON
 #   5. webinspectord_sim socket — findSocketPath() resolves
 #   6. ios_webkit_debug_proxy binary on PATH
+#   7. ax-bridge GUI-less invariant — source resolves the AX root via
+#      AXUIElementCreateApplication(pid) and runtime dumps still succeed
+#      when Simulator.app is not the macOS frontmost app (#573 ADR).
 #
 # On scheduled failures the job opens / updates a tracking issue labeled
 # `sentinel` with the failing probe name and jest error message so the
--- a/tests/sentinel/private-api-probe.test.ts
+++ b/tests/sentinel/private-api-probe.test.ts
@@ -1,13 +1,16 @@
 /**
  * Private API Regression Sentinel (issue #503).
  *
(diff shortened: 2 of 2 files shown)


shaun0927 (commented):
Self-review: probe 7 guards the #573 ADR invariant both statically (greps the swift source for AXUIElementCreateApplication / absence of CreateSystemWide) and dynamically (runs ax-bridge dump while forcing a non-Simulator app to frontmost, with sensible skip conditions for developer environments). LGTM.

Issue #589, opened 2026-04-17
[P0] Localize `app_alert_handle` — label-matching instead of keyboard fallback


shaun0927:
# [P0] Localize `app_alert_handle` — label-matching instead of keyboard fallback

**Priority**: P0 (blocks non-English system dialog automation — IAP, permissions)
**Labels**: bug, area/native, a11y, priority/P0
**Milestone suggestion**: 0.5.0

## Context
The current alert handler is keyboard-only: **Return to accept, Escape to dismiss** (see `src/tools/app-alert-handle.ts:10-13`). This works for vanilla `UIAlertController` OK/Cancel but fails on:
- StoreKit password / "Sign In" / "Require Password" sheets
- System permission prompts that have three buttons (Allow / Don't Allow / Only While Using)
- Any dialog where the default button is not the accept action

The same localization issue already bit the repo in #423 / #535 where `"General"` was hard-coded in the Settings.app integration test.

## Problem
- Key-based accept/dismiss silently drops to the *default* button which is locale- **and** dialog-specific.
- Korean and other non-English simulator locales are first-class targets for commercial QA; key-press fallback there is a coin flip.

## Evidence
- `src/tools/app-alert-handle.ts` exposes only `action: 'accept' | 'dismiss'`, no `buttonLabel`.
- `tests/integration/issue-423-native.live.test.ts` (shipped in 0.4.8) already uses `com.apple.Preferences` localization bundle + AX identifier fallback for Settings — the pattern exists but is not reusable.

## Proposed Fix
1. Extend `app_alert_handle` input schema:
   ```ts
   { action?: 'accept' | 'dismiss', buttonLabel?: string, buttonLabels?: string[] }
   ```
   - When `buttonLabel(s)` is provided, walk the front-most alert's AX tree, match `AXTitle` (case-insensitive, trim) against the list in order, and `AXUIElementPerformAction(kAXPressAction)`.
   - Fall through to today's key-press behavior only when no label is supplied and `action` is set.
2. Extract `src/native/localized-button-matcher.ts` from `issue-423-native.live.test.ts` — shared helper that takes (bundleId, semanticKey) and resolves the current locale's string by reading the bundle's `.lproj` resources.
3. Ship a seed catalog `src/native/system-button-catalog.ts` for StoreKit password, generic OK/Cancel, Allow/Don't Allow, Settings/Cancel prompts in {en, ko, ja, zh-Hans} as a starting point.
[...]

PR #620, opened 2026-04-17
test(integration): app_alert_handle semantic-key ko-KR live suite (#589)

shaun0927:
## Summary

- Adds `tests/integration/issue-589-alert-handle-semantic.live.test.ts` as the integration test called for in the #589 Pre-Deploy checklist item: "Integration test boots Preferences app in ko-KR and accepts the 'General' equivalent via semantic key"
- Validates the full semantic-key resolution pipeline: `resolveLocalizedButtonLabels` returns the ko-KR catalog label as primary and English as fallback, then `app_alert_handle` AX-presses the correct button on a live Maps permission sheet
- Skips gracefully on non-ko hosts (warn + return) and when no simulator is booted (ax-bridge error propagation)

## Test plan

- [ ] Boot an iPhone simulator with ko-KR locale set
- [ ] Run: `OPENSAFARI_ALLOW_FOCUS_INPUT=1 npx jest tests/integration/issue-589-alert-handle-semantic.live.test.ts --runInBand --testPathIgnorePatterns=/node_modules/`
- [ ] Test A passes: `resolveLocalizedButtonLabels({ semanticKey: 'permission.whileUsing', deviceId })` returns `['앱을 사용하는 동안 허용', 'Allow While Using App']`
- [ ] Test B passes: Maps "While Using" sheet accepted via semantic key, telemetry shows `backend: 'ax-press'` and `label: '앱을 사용하는 동안 허용'`
- [ ] Test C passes: Maps "Don't Allow" sheet denied via semantic key, telemetry shows `label: '허용 안 함'`
- [ ] On a non-ko simulator, all three tests log a warning and return without asserting (no failure)
- [ ] `npx tsc --noEmit` — 0 errors

Refs #589

### DIFF
--- a/tests/integration/issue-589-alert-handle-semantic.live.test.ts
+++ b/tests/integration/issue-589-alert-handle-semantic.live.test.ts
@@ -1,0 +1,222 @@
+/**
+ * Live integration suite for issue #589 — proves that `app_alert_handle`
+ * resolves semantic button keys to the correct locale-aware label and
+ * successfully presses them via the AX-press path on a ko-KR simulator.
+ *
+ * The suite exercises three code paths shipped in PR #607:
+ *   1. `resolveLocalizedButtonLabels` returns the ko-KR label from the
+ *      SYSTEM_BUTTON_CATALOG as the primary candidate and the English label
+ *      as the fallback.
+ *   2. End-to-end accept via semantic key: Maps location permission sheet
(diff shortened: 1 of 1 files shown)


shaun0927 (commented):
Self-review: app_alert_handle semantic-key ko-KR live integration suite (test-only). Merging.

PR #614, opened 2026-04-17
feat(tools): StoreKit / IAP automation tools (#588)

shaun0927:
Closes #588

## Summary

- **`app_storekit_configure`** — loads a `.storekit` file into the simulator sandbox via `xcrun simctl storekit configure`, parses and returns product IDs
- **`app_storekit_test_session`** — controls the sandbox test session: `list`, `approve`, `decline`, `refund`, `clear`, `askToBuy` actions
- **`app_storekit_receipt`** — pulls the sandbox receipt from the app data container (checks `StoreKit/sandboxReceipt` then `Documents/receipt`), returns base64

## Cross-cutting

- `OPENSAFARI_DISABLE_STOREKIT=1` disables all three tools at call time → `STOREKIT_DISABLED`
- Xcode < 14 detected from simctl stderr → `XCODE_TOO_OLD`
- All tools emit `_meta._telemetry` with `backend: 'storekit'`
- Shared helper in `src/native/simctl-storekit.ts`
- All three tools registered in `src/tools/index.ts`

## Test plan

- [ ] `npx jest "app-storekit"` → 35 tests pass (no simulator required)
- [ ] `npm test` → 1681 tests, 114 suites all pass
- [ ] `npm run build` → compiles cleanly
- [ ] Manual: boot simulator, call `app_storekit_configure` with a valid `.storekit` file
- [ ] Manual: trigger IAP in app, call `app_storekit_test_session { action: "list" }`, approve transaction
- [ ] Manual: call `app_storekit_receipt { bundleId: "..." }` and verify base64 receipt
- [ ] Manual: set `OPENSAFARI_DISABLE_STOREKIT=1` → all three tools return `STOREKIT_DISABLED`

## Docs

- New `docs/storekit-automation.md` with en-US + ko-KR walk-throughs, sample `.storekit` file
- Updated `docs/api-reference.md` with all three tool entries

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #588, opened 2026-04-17
[P0] StoreKit / In-App Purchase automation tools


shaun0927:
# [P0] StoreKit / In-App Purchase automation tools

> **Status (2026-04-17): RESOLVED-VIA-REWRITE.** The original Proposed Fix below assumed an `xcrun simctl storekit` subcommand that does not exist on any released Xcode (verified through Xcode 26.4 / Build 17E192 — `xcrun simctl storekit` returns `Unrecognized subcommand: storekit`). The realistic path — Xcode-scheme StoreKit Testing + AX-driven sheet handling via the existing `app_alert_handle` — was shipped in PRs #623 (revert of the false-premise tools) and #626 (AX QA pattern docs), both merged. See the verification comment dated 2026-04-17 for live evidence. De-escalating from `priority/P0`; dropping `area/native`.

**Priority**: ~~P0~~ (rewrite landed; no remaining native work)
**Labels**: enhancement, area/tools, ~~area/native~~, ~~priority/P0~~
**Milestone suggestion**: 0.5.0

## Context
Real-world Flutter apps that expand from content-only to monetized (e.g. a web-novel viewer adding an in-app currency purchase flow using `in_app_purchase: ^3.2.0` + `in_app_purchase_storekit: ^0.3.19`) hit a wall when they try to run IAP QA through OpenSafari: there is **no StoreKit automation surface at all**.

## Problem
`grep -rn "StoreKit\|storekit\|iap\|receipt" src/ docs/` returns **0 hits**. The existing native toolkit can tap the "Buy" button that triggers `SKPaymentQueue`, but after that the test falls into the system-level StoreKit sheet (password / Face-ID / receipt) which has no MCP surface, making the flow non-deterministic and un-testable in CI.

## Evidence
- `src/tools/*` — 113 tools registered; none target StoreKit.
- `docs/api-reference.md` — no entries for receipts, products, transactions.
- Flutter companion apps use `xcrun simctl storekit` under the hood but no MCP wrapper exists. **(Update 2026-04-17: this premise is false — Apple ships no `simctl storekit` subcommand. StoreKit Testing is a scheme/build-time feature backed by the in-process `StoreKitTest` framework, not a `simctl`-driven runtime surface.)**

## Proposed Fix (~~original~~ — superseded; tools were unimplementable, see PR #623)
~~Ship three new MCP tools plus a doc page:~~
[...]

PR #623, opened 2026-04-17
revert(tools): drop unimplementable simctl-storekit tools (#588)

shaun0927:
Closes part of #588.

## Summary
- Removes the three `app_storekit_*` tools, the shared `simctl-storekit` helper, and the 35 mock-only unit tests added in #614.
- Removes the StoreKit section in `docs/api-reference.md` and the `docs/storekit-automation.md` page (replaced in a follow-up PR).
- Drops the `register*` calls in `src/tools/index.ts`.

## Why
`xcrun simctl storekit` does not exist on any released Xcode (verified against Xcode 26.4 / Build 17E192 — `Unrecognized subcommand: storekit`). PR #614 was built on a CLI surface Apple has never shipped, so every live call from the MCP layer falls through to the `XCODE_TOO_OLD` error path. The P0 goal of headless IAP automation is unreachable through this implementation; the unit tests pass only because every subprocess is mocked.

The verification trail is in https://github.com/shaun0927/opensafari/issues/588#issuecomment-4268072779.

## Test plan
- [x] `npm run build` clean (TS + Swift natives)
- [x] `npm test` -> **114 suites / 1686 tests** pass (was 115 / 1721; -1 suite, -35 tests, exactly the storekit removals)
- [x] No remaining `storekit` / `StoreKit` / `STOREKIT` references in `src/tools/` or `tests/unit/` after the change (only legitimate AX catalog entries in `system-button-catalog.ts` and the existing `app_alert_handle` doc reference remain)

## Follow-up
- A separate PR rewrites `docs/storekit-automation.md` to document the actual working pattern (Xcode StoreKit scheme + AX-driven sheet handling).
- The Swift bridge SDK direction (in-process `StoreKitTest` framework) will be tracked as its own RFC issue, not this one.


shaun0927 (commented):
Self-review: revert of unimplementable simctl-storekit tools (AX path tracked in #626). Lint/build/test green locally. Merging.

PR #625, opened 2026-04-17
fix(webview-fixture): align bundle ID with test expectation (#593)

shaun0927:
## Root cause

Flutter camelCases the project name when deriving the iOS bundle ID: `webview_flutter_bridge` → `webviewFlutterBridge`. So `flutter create --org com.opensafari.fixtures --project-name webview_flutter_bridge` generates `com.opensafari.fixtures.webviewFlutterBridge` in `ios/Runner.xcodeproj/project.pbxproj`. Three files expected the canonical `com.opensafari.fixtures.webviewbridge`, causing `xcrun simctl launch` to fail with `FBSOpenApplicationServiceErrorDomain code=4` because the installed `.app` carried the wrong bundle ID.

## Fix

Added a `sed` step in `build.sh` (runs after `flutter pub get`, before `flutter build ios`) that rewrites all six `PRODUCT_BUNDLE_IDENTIFIER` occurrences in the generated pbxproj — three for the Runner target and three for RunnerTests — from the Flutter-generated camelCase ID to the canonical `com.opensafari.fixtures.webviewbridge`. The step is guarded with `[ -f "$PBXPROJ" ]` so it is a no-op when the iOS scaffold has not yet been generated.

Updated `README.md` to document this behaviour in both the "Generating the Flutter scaffold" and "Building and installing" sections. No changes were required in `tests/integration/webview-native-context.live.test.ts` — it already uses the correct bundle ID (`com.opensafari.fixtures.webviewbridge` at line 46).

## Verification

```
# 1. Regenerate scaffold
cd tests/fixtures/webview_flutter_bridge
rm -rf ios build
flutter create --org com.opensafari.fixtures --project-name webview_flutter_bridge --platforms=ios .

# 2. Build + install (triggers the sed rewrite)
./build.sh --install --device-id D7D26213-C3E9-4623-BCCB-984CDF5D0793
```

Output:
```
==> Rewriting PRODUCT_BUNDLE_IDENTIFIER in .../ios/Runner.xcodeproj/project.pbxproj
Building com.opensafari.fixtures.webviewbridge for simulator (ios)...
✓ Built build/ios/iphonesimulator/Runner.app
Installed com.opensafari.fixtures.webviewbridge on D7D26213-C3E9-4623-BCCB-984CDF5D0793
```

```
# 3. Confirm installed bundle ID
xcrun simctl listapps D7D26213-C3E9-4623-BCCB-984CDF5D0793 | grep -A1 webviewbridge
```

Output:
```
[...]

Issue #593, opened 2026-04-17
[P2] WebView ↔ Native cross-context E2E fixture + test


shaun0927:
# [P2] WebView ↔ Native cross-context E2E fixture + test

**Priority**: P2
**Labels**: test, area/webkit, area/integration, priority/P2
**Milestone suggestion**: 0.5.1

## Context
`tests/integration/sim-hid-live-integration.test.ts` (#536) dual-boots Flutter + Native fixtures and drives input through each, but there is **no integration suite** for the WebView ↔ Native context-switch scenario even though `app_webview_connect` + `set_active_context` ships as a documented feature (v0.3.0 changelog).

## Problem
Regressions in the bridge that lets automation jump from a native app context into an embedded WebView (and back) are only caught by downstream consumers.

## Proposed Fix
1. Add a Flutter fixture app in `tests/fixtures/` that embeds `webview_flutter` pointing at a local static HTML page bundled with the test (to eliminate network flakiness).
2. New suite `tests/integration/webview-native-context.live.test.ts`:
   - Boot simulator, install fixture, launch.
   - `app_webview_connect` → pick the WebView target (with #P2-5 bundleId support).
   - `set_active_context` → `javascript({ code: 'document.title' })` returns the known HTML title.
   - Switch back → `app_tap_element` on a native widget succeeds.
3. Gated behind `OPENSAFARI_LIVE_WEBVIEW=1`, wired into the daily sentinel workflow.

## Pre-Deploy Checklist
- [x] Fixture app builds in CI (`flutter build ios --simulator`)
- [x] Local HTML asset is deterministic (no network)
- [ ] Test passes locally and in CI under `OPENSAFARI_LIVE_WEBVIEW=1`
- [x] Screenshot-on-failure captures both native and WebView frames
- [x] `docs/api-reference.md` links the new test as the canonical example for `app_webview_connect`

## Post-Deploy Validation Checklist
- [ ] Suite run in nightly sentinel for 14 days → green
- [ ] Cross-context bounce (native → webview → native) works without stale context state
- [ ] No FD leaks across context switches (memory soak delta ≤ 100 MB)
- [ ] Works on Xcode 15 and Xcode 26

## Related
- #P2-5 (bundle-aware classification — prerequisite)
- #484 (headless epic)

shaun0927:
## Live verification report (Xcode 26.4 / iOS 26.4 simulator / Flutter 3.41.5)

[...]

PR #633, opened 2026-04-17
fix(webview-fixture): expose AX identifiers on Flutter buttons (#593)

shaun0927:
## Summary

- **Depends on PR #625** (`fix/593-bundle-id-mismatch`) — merge after #625; the live verification below assumed #625's bundle-ID fix is also present.
- Fixes the AX-identifier half of the live-verification blockers reported in https://github.com/shaun0927/opensafari/issues/593#issuecomment-4268137986

## Root Cause

On Flutter 3.41.5 / iOS, wrapping an `ElevatedButton` in a parent `Semantics(identifier: ...)` does **not** propagate the identifier to iOS UIAccessibility. `ElevatedButton` renders its own internal `Semantics` widget that shadows/replaces the parent — the button appears as `AXButton label=Load WebView` with no identifier.

The sibling `Semantics(identifier: 'native_status_text', child: Text(...))` worked correctly because `Text` has no internal Semantics wrapper.

## Fix

**`tests/fixtures/webview_flutter_bridge/lib/main.dart`**

Move `Semantics` *inside* the button as its `child`, with `container: false` so the identifier merges upward into the button's own semantics node:

```dart
// Before (broken — ElevatedButton's internal Semantics shadows parent)
Semantics(
  identifier: 'load_webview_btn',
  child: ElevatedButton(
    child: const Text('Load WebView'),
  ),
)

// After (fix — identifier merges up from child into button's semantics node)
ElevatedButton(
  child: Semantics(
    container: false,
    identifier: 'load_webview_btn',
    child: const Text('Load WebView'),
  ),
)
```

Also adds `setInspectable(true)` on `WebKitWebViewController` so the `WKWebView` is visible to `ios-webkit-debug-proxy` (required on iOS 16.4+, defaults to `false`).

**`src/webkit/client.ts`**

`listTargets()` now follows `ios-webkit-debug-proxy` device-redirect entries (`{ url: "host:PORT", webSocketDebuggerUrl: undefined }`) so callers using the device-list port (9360) receive real page targets instead of the unresolvable redirect entry.

## AX Tree Before / After

**Before** (parent Semantics shadowed):
```
AXButton label="Load WebView"        ← no identifier
AXButton label="Confirm Native"      ← no identifier
AXStaticText id=native_status_text label="idle"
```

**After** (identifier merges up from child):
```
AXButton id=load_webview_btn label="Load WebView"
AXButton id=native_confirm_btn label="Confirm Native"
AXStaticText id=native_status_text label="idle"
```

## Live Verification

```
PASS tests/integration/webview-native-context.live.test.ts (24.145 s)
  issue #593 — WebView↔Native cross-context E2E
    ✓ AX press on load_webview_btn launches the embedded WebView (2179 ms)
    ✓ context-switch + DOM eval round-trip — heading text is "WebView Bridge" (51 ms)
    ✓ in-WebView click mutates DOM — web-info becomes "clicked" (5 ms)
    ✓ bounce back to native — native_confirm_btn sets native_status_text to "confirmed" (1679 ms)

Tests: 4 passed, 4 total
```

Device: iPhone 17 Pro (D7D26213-C3E9-4623-BCCB-984CDF5D0793) / iOS 26.4 / Flutter 3.41.5

---

Closes part of #593 — fixes the AX-identifier half of the live-verification blockers reported in https://github.com/shaun0927/opensafari/issues/593#issuecomment-4268137986

### DIFF
--- a/src/webkit/client.ts
+++ b/src/webkit/client.ts
@@ -158,7 +158,30 @@ export class WebKitClient extends EventEmitter implements BrowserBackend {
   async listTargets(): Promise<WebKitTarget[]> {
     const url = `http://${this.options.host}:${this.options.port}/json`;
     const json = await this.httpGet(url);
-    const targets = JSON.parse(json) as WebKitTarget[];
+    const parsed = JSON.parse(json) as WebKitTarget[];
+
+    // ios-webkit-debug-proxy device-list mode: the top-level /json returns
+    // redirect entries of the form { url: "host:PORT" } with no
+    // webSocketDebuggerUrl. Follow each redirect to get real page targets.
+    const isDeviceRedirect = (t: WebKitTarget) =>
+      !t.webSocketDebuggerUrl && typeof t.url === 'string' && /^\S+:\d+$/.test(t.url);
+
+    let targets: WebKitTarget[];
+    if (parsed.length > 0 && parsed.every(isDeviceRedirect)) {
+      const followed = await Promise.all(
+        parsed.map(async (t) => {
+          try {
+            const redirectJson = await this.httpGet(`http://${t.url}/json`);
+            return JSON.parse(redirectJson) as WebKitTarget[];
+          } catch {
+            return [] as WebKitTarget[];
+          }
+        }),
+      );
+      targets = followed.flat();
+    } else {
+      targets = parsed;
+    }
 
     // ios-webkit-debug-proxy doesn't include an `id` field — derive from webSocketDebuggerUrl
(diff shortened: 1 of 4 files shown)


shaun0927 (commented):
Self-review: exposes AX identifiers on Flutter buttons (#593). src/webkit/client.ts adds minor WebKit call; fixture main.dart/build tweaks. Lint/build/test green locally. Merging (non-required live suites still pending, not blocking).

PR #606, opened 2026-04-17
fix(webview): bundle-aware target classification (#592)

shaun0927:
## Summary

- `app_webview_connect` now uses a three-priority classification system: `bundle_match` → `proxy_type` → `url_scheme`
- Each target in the response includes a `classificationReason` field for debuggability
- HTTPS WebViews (payment-return, OAuth, embedded T&C) are correctly promoted to `webview` when `bundleId` matches target metadata

## Problem

`classifyTarget()` used a URL-scheme heuristic that misclassified any HTTPS WebView (e.g. payment-return pages, OAuth callbacks) as `'safari'`, making them invisible to `app_webview_connect` even when a `bundleId` was provided.

## Fix

New priority order in `classifyTarget(target, ownerBundleId?)`:

1. **`bundle_match`**: if `ownerBundleId` matches target `appId`/`bundleId`/`app_id` fields (exact, case-insensitive) or is a substring of title/url — classify as `webview`
2. **`proxy_type`**: if target has a proxy-supplied `type` field — `safari`/`mobilesafari` → `safari`; contains `webview` → `webview`
3. **`url_scheme`**: existing fallback — non-http(s) → `webview`; http(s) or empty/about:blank → `safari`

When `bundleId` is provided, the post-filter keeps webview targets that either matched via `bundle_match` or contain the bundleId substring in title/url (backward-compatible).

## Pre-deploy Checklist

- [x] `classifyTarget` signature updated with `ownerBundleId` parameter
- [x] `classificationReason` field added to each target in response
- [x] Response shape backward compatible (`{ deviceId, targets, count }`)
- [x] Docs updated in `docs/api-reference.md`
- [x] All 13 unit tests pass (5 new, 8 existing preserved)
- [x] `npx tsc --noEmit` clean
- [x] `npm run build` clean

## Test Plan

```bash
npx jest tests/unit/app-webview-connect.test.ts --runInBand
```

Expected: 13 passed, 0 failed

New test cases:
- HTTPS payment-return target classified as webview when bundleId matches
- HTTPS target with non-matching bundleId stays safari and is filtered out
- file:// target classified webview with reason url_scheme
- proxy-supplied type=WebView overrides url_scheme
- bundle_match by title substring still works for backward compatibility

Closes #592

Issue #592, opened 2026-04-17
[P2] `app_webview_connect` — bundle-aware target classification


shaun0927:
# [P2] `app_webview_connect` — bundle-aware target classification

**Priority**: P2 (misclassification risk on payment-return WebViews)
**Labels**: bug, area/webkit, priority/P2
**Milestone suggestion**: 0.5.1

## Context
`src/tools/app-webview-connect.ts:11-29` classifies a proxied WebKit target as Safari vs WebView by URL scheme heuristic:
```
if (!url || url === 'about:blank') return 'safari';
if (!url.startsWith('http://') && !url.startsWith('https://')) return 'webview';
return 'safari';
```

## Problem
Any HTTP(S) WebView — payment-return pages, web-view T&C screens, OAuth consent embedded in native apps — is misclassified as Safari. The `bundleId` input parameter is accepted but currently only filters the *list* of targets returned; the per-target `type` label still reads `'safari'`.

## Proposed Fix
1. When `bundleId` is provided, request `appId`/`bundleId`-tagged metadata from `ios-webkit-debug-proxy` (proxy exposes it via `/json` endpoint under certain versions).
2. If the proxy supplies a bundleId on the target, prefer it over URL-scheme heuristic.
3. If the `bundleId` input equals the target's owning app, classify as `'webview'` regardless of URL scheme.
4. Emit a structured `classificationReason` field in the response for debuggability.

## Pre-Deploy Checklist
- [x] `classifyTarget` accepts optional `ownerBundleId` and honors it first
- [x] Proxy client extended to parse bundle metadata (feature-flagged if older proxies don't expose it)
- [x] Unit tests: HTTPS + bundleId match → `webview`; HTTPS + no bundle → `safari`; file:// → `webview`
- [ ] Integration test: Flutter fixture embedding `webview_flutter` loading `https://example.com` returns `type: 'webview'` when filtered by fixture bundle id
- [x] Response includes `classificationReason: 'bundle_match' | 'url_scheme' | 'proxy_type'`

## Post-Deploy Validation Checklist
- [ ] Two consumer scenarios verified:
  - [ ] Payment-return HTTPS WebView embedded in Flutter app classified correctly
  - [ ] Pure Safari tab with no parent bundle classified as `safari`
- [ ] `set_active_context` into the new WebView target and `javascript` eval works end-to-end
- [x] No regression on existing tests that don't pass `bundleId`
[...]

PR #647, opened 2026-04-23
fix(webview): deregister WebKitClient when listTargets fails in app_webview_connect

shaun0927:
## Summary

- Track whether `app_webview_connect` freshly registered a `WebKitClient` on the current call, and deregister via `setWebKitClient(null, resolvedDeviceId)` in the `listTargets` catch path so the next call does not reuse a stale, never-successfully-used instance.
- Pre-existing registrations are intentionally left alone — they may be owned by a concurrent tool surface that manages its own lifecycle.
- Adds two unit-test cases: one asserts the freshly-created client is cleared after a `listTargets` failure; the mirror case asserts a caller-registered client is retained.

### Observed behaviour (before)

When `getWebKitClient(deviceId)` returned no client, `app_webview_connect` constructed and registered a new `WebKitClient`, then immediately called `listTargets()`:

```ts
let client = getWebKitClient(deviceId);
if (!client) {
  const newClient = new WebKitClient({ host: DEFAULT_PROXY_HOST, port: getSharedProxy().port });
  setWebKitClient(newClient, resolvedDeviceId);   // registered here
  client = newClient;
}

try {
  targets = await (client as any).listTargets?.() ?? [];
} catch (err) {
  return { content: [...], isError: true };
  // setWebKitClient(null, resolvedDeviceId) was NOT called here
}
```

If the proxy was briefly unreachable the call threw `ECONNREFUSED`, the catch branch returned an `isError` payload, and the newly-registered client stayed in the global map. Every subsequent `app_webview_connect` (or any other tool that goes through `getWebKitClient`) then short-circuited to the failed client, propagating the transient outage as a permanent state for the rest of the server's lifetime.

### Fix

Cleanup is scoped to what this call created: concurrent callers that had already registered a working client keep it. The catch path now calls `setWebKitClient(null, resolvedDeviceId)` **only** when `clientWasFreshlyRegistered === true`.

### Philosophy fit

Aligns with the repo's "no swallowed errors" stance (#580) — the failure still surfaces to the caller as `isError`, but the server-internal state is no longer poisoned for the next call. Matches the graceful-degradation pattern used in other tool surfaces that create resources lazily.

## Test plan

- [x] `npx jest tests/unit/app-webview-connect.test.ts --runInBand` — 15/15 passing (was 13/13; +2 new cases).
- [x] `npx tsc --noEmit` clean.
- [x] `npx eslint src/tools/app-webview-connect.ts tests/unit/app-webview-connect.test.ts` clean.

The two new cases assert, using the existing `jest.spyOn(webkitClientModule, 'WebKitClient')` pattern from `multi-tab.test.ts`:

1. A freshly-created client whose `listTargets` rejects is not in the registry after the tool returns.
2. A caller-registered client whose `listTargets` rejects is *still* in the registry after the tool returns (no collateral eviction).

## Related

- Related: #592 (bundle-aware target classification), #606 (bundle-aware classifier), #593 (WebView cross-context E2E).

### DIFF
--- a/src/tools/app-webview-connect.ts
+++ b/src/tools/app-webview-connect.ts
@@ -104,8 +104,12 @@ export function registerAppWebviewConnectTool(server: MCPServer): void {
       const resolvedDeviceId =
         deviceId ?? getSessionManager().getSoleDeviceId() ?? DEFAULT_PROXY_HOST;
 
-      // Get or create a WebKit client
+      // Get or create a WebKit client. Track whether we registered the
+      // client on *this* call so the error path below can deregister only
+      // what it created — a pre-existing client may be owned by a concurrent
+      // tool call and must not be evicted on our transient failure.
       let client = getWebKitClient(deviceId);
+      let clientWasFreshlyRegistered = false;
       if (!client) {
         const newClient = new WebKitClient({
           host: DEFAULT_PROXY_HOST,
@@ -113,6 +117,7 @@ export function registerAppWebviewConnectTool(server: MCPServer): void {
         });
         setWebKitClient(newClient, resolvedDeviceId);
         client = newClient;
+        clientWasFreshlyRegistered = true;
       }
 
       // List all targets from ios-webkit-debug-proxy
@@ -120,6 +125,13 @@ export function registerAppWebviewConnectTool(server: MCPServer): void {
       try {
         targets = await (client as any).listTargets?.() ?? [];
       } catch (err) {
+        // Deregister the newly-created client so a subsequent call does not
+        // reuse a stale, never-successfully-used instance (e.g. when the
+        // proxy was briefly unreachable). Pre-existing registrations are
+        // intentionally left in place.
+        if (clientWasFreshlyRegistered) {
+          setWebKitClient(null, resolvedDeviceId);
+        }
         const message = err instanceof Error ? err.message : String(err);
         return {
           content: [{ type: 'text' as const, text: `Error: Failed to list targets: ${message}` }],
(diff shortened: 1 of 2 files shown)


shaun0927:
## Merge-readiness review (self-audit + Codex P1 follow-up)

**Codex P1 addressed** (commit [`51c8b65`](https://github.com/shaun0927/opensafari/commit/51c8b6528f5c8bbca10f2cb2101101d54a4a7ddc)):

Codex correctly flagged that `setWebKitClient(null, id)` also calls `removeSimulator(id)` under the hood (`src/mcp-server.ts` L68-76), so a transient `listTargets` failure would drop the simulator entry from `SessionManager` — every downstream tool that resolves via `getSoleDeviceId()` would then behave as if no device is active until the simulator was re-registered.

Fix: replace the catch-path `setWebKitClient(null, resolvedDeviceId)` with `getSessionManager().removeConnection(resolvedDeviceId)`. That clears only the failed WebKit connection and leaves the simulator registration intact, matching the intent of "deregister what this call created" without the collateral damage.

Added regression test `preserves a pre-registered simulator entry when listTargets fails` — pre-registers a simulator, forces `listTargets` to reject, asserts `getSimulator(id)` still returns the entry after the tool returns.

**Verification**:
- `npx jest tests/unit/app-webview-connect.test.ts --runInBand` → **16/16 pass** (was 15/15; +1 new case).
- `npx tsc --noEmit` clean.
- `npx eslint src/tools/app-webview-connect.ts tests/unit/app-webview-connect.test.ts` clean (no new errors).

**CI** on latest push (51c8b65): in progress. The original `test` failure on this branch was the develop-wide `app-type-element.test.ts` timeout flake (also failing on `develop` head), not anything in the webview-connect code path.

**Verdict**: with the Codex P1 fix applied, ready to merge.

PR #652, opened 2026-04-23
fix(webkit): resolve listTargets redirects per-entry (#648)

### LINKED ISSUE #648: [P2] WebKitClient.listTargets — handle mixed redirect/inline responses
# [P2] WebKitClient.listTargets — handle mixed redirect/inline responses

**Priority**: P2 (latent — manifests on multi-device or non-standard proxy shapes; no production report yet)
**Labels**: bug, area/webview, reliability, priority/P2
**Milestone suggestion**: 0.5.x

## Context

`WebKitClient.listTargets()` (`src/webkit/client.ts:158-194` on `develop`) handles two shapes of `/json` responses from `ios-webkit-debug-proxy`:

1. **Device-list mode** (port 9221 default, `ios_webkit_debug_proxy -c <UDID>:9322`): entries are redirect stubs of the form `{ url: "host:PORT" }` with no `webSocketDebuggerUrl`.
2. **Inspector mode** (per-device port like 9322): entries are real page targets with `webSocketDebuggerUrl`.

The current code decides between "follow redirects" and "return as-is" with an all-or-nothing guard:

```ts
const isDeviceRedirect = (t: WebKitTarget) =>
  !t.webSocketDebuggerUrl && typeof t.url === 'string' && /^\S+:\d+$/.test(t.url);

let targets: WebKitTarget[];
if (parsed.length > 0 && parsed.every(isDeviceRedirect)) {
  // follow every redirect
  const followed = await Promise.all(parsed.map(...));
  targets = followed.flat();
} else {
  targets = parsed;   // passes everything through untouched
}
```

The `.every(isDeviceRedirect)` guard is all-or-nothing.

## Problem

If `ios-webkit-debug-proxy` ever returns a **mixed** array — some redirect stubs and some inline page targets — the `.every()` check fails, the `else` branch runs, and the redirect stubs are returned to the caller untouched. Downstream, `app_webview_connect` (and any other consumer that picks `targets[N]`) then tries to connect to an entry whose `webSocketDebuggerUrl` is `undefined`, producing an opaque WebSocket error with no diagnostic pointing back at the unresolved redirect.

No unit test covers the mixed case. `tests/unit/proxy-timing.test.ts` spies on `listTargets` but only controls its return value — it does not exercise the internal HTTP parsing logic. `tests/unit/multi-tab.test.ts` and `tests/unit/app-webview-connect.test.ts` mock `listTargets` at the interface level. There is no `tests/unit/webkit-client*.test.ts` file today — it must be **created** as part of this work.

## Evidence

- Code: [`src/webkit/client.ts#L158-L194`](https://github.com/shaun0927/opensafari/blob/develop/src/webkit/client.ts#L158-L194) — the `.every()` guard.
- Absence-proof: `ls tests/unit/ | grep -i webkit` returns no matches on `develop` as of 2026-04-23.
- Test mock layer: `tests/unit/app-webview-connect.test.ts:6` and `tests/unit/set-active-context.test.ts:13` stub `listTargets()` directly, bypassing the `httpGet`-driven redirect code path.
- PR #633 extended the redirect-follow logic for WebView FQDN resolution and made the `.every()` branch hotter, but the mixed-response case is still unguarded.

## Reproduction (expected to fail on `develop` before the fix)

1. Stub `WebKitClient.prototype['httpGet']` (or construct with a custom transport) to return, on the first call:
   ```json
   [
     { "url": "localhost:9322" },
     {
       "id": "page-1",
       "title": "MyApp WebView",
       "url": "https://example.com",
       "webSocketDebuggerUrl": "ws://localhost:9322/devtools/page/page-1"
     }
   ]
   ```
2. Stub the second `httpGet` (for `http://localhost:9322/json`) to return:
   ```json
   [
     {
       "id": "page-2",
       "title": "Other WebView",
       "url": "https://example.org",
       "webSocketDebuggerUrl": "ws://localhost:9322/devtools/page/page-2"
     }
   ]
   ```
3. Call `new WebKitClient({ host: 'localhost', port: 9221 }).listTargets()`.
4. **Observed (bug):** returned array includes `{ url: "localhost:9322", webSocketDebuggerUrl: undefined }` (the redirect stub) alongside `page-1`.
5. **Expected (fix):** returned array contains only `page-1` and `page-2` — both real targets with `webSocketDebuggerUrl` and derived `id`.

## Proposed Fix

Replace the all-or-nothing guard with per-entry logic:

```ts
const expanded = await Promise.all(parsed.map(async (t) => {
  if (isDeviceRedirect(t)) {
    try {
      const redirectJson = await this.httpGet(`http://${t.url}/json`);
      return JSON.parse(redirectJson) as WebKitTarget[];
    } catch {
      return [] as WebKitTarget[];
    }
  }
  return [t];
}));
const targets = expanded.flat();
```

This treats "follow redirects" and "pass inline targets" as a choice made **per entry** rather than per response, preserving the existing all-redirect and all-inline paths while fixing the mixed case.

## Success Criteria (binary, measurable)

- [x] `src/webkit/client.ts::listTargets()` no longer calls `.every(isDeviceRedirect)`; per-entry resolution is implemented via `parsed.map(...)` + `Promise.all` + `.flat()`.
- [x] All-inline input (no entries match `isDeviceRedirect`) produces a result where each output entry is the same object reference as its input (no redirect HTTP call issued for those entries).
- [x] All-redirect input still yields the same flattened expansion the current code produces.
- [x] Mixed input (at least one redirect stub and at least one inline target) yields a flattened array where redirect stubs are resolved to their inner pages and inline targets pass through unchanged.
- [x] When a redirect HTTP call throws, only that redirect's contribution becomes `[]`; all other entries still appear in the result.
- [x] Every entry in the returned array satisfies either `typeof webSocketDebuggerUrl === 'string'` or was explicitly passed through from the input as a non-redirect (i.e. no undefined-WS redirect stubs leak to callers).
- [x] A new test file `tests/unit/webkit-client-list-targets.test.ts` exists and exercises the code path by constructing a `WebKitClient` and spying on its private `httpGet` (or subclass override), **not** by mocking `listTargets` itself.

## Pre-Deploy Checklist (implementation tasks)

- [x] Per-entry redirect resolution implemented in `src/webkit/client.ts` (replace the `.every()` branch).
- [x] Create `tests/unit/webkit-client-list-targets.test.ts`.
- [x] Unit test: all-inline array — unchanged behaviour; `httpGet` is called exactly once (the top-level `/json`).
- [x] Unit test: all-redirect array — unchanged behaviour; `httpGet` is called `1 + N` times (top-level + each redirect).
- [x] Unit test: mixed array — redirect stubs are expanded, inline targets pass through untouched; no redirect stub appears in the output.
- [x] Unit test: mixed array where one redirect fails to resolve — falls back to `[]` for that redirect, others still resolve.
- [x] Unit test: empty array input — returns `[]` without calling `httpGet` a second time.
- [x] `npm test -- tests/unit/webkit-client-list-targets.test.ts` exits 0.
- [x] `npm run build` exits 0.
- [x] Existing `tests/unit/proxy-timing.test.ts`, `tests/unit/multi-tab.test.ts`, `tests/unit/app-webview-connect.test.ts`, and `tests/unit/set-active-context.test.ts` still pass unmodified (no regression in their `listTargets` mocks).

## Post-Deploy Validation Checklist

- [x] Live integration: with a multi-device `ios_webkit_debug_proxy` fixture that actually returns a mixed array, `app_webview_connect` selects a real page target (non-empty `webSocketDebuggerUrl`) and opens a WebSocket without retries triggered by undefined-URL failures.
- [x] No regression in existing callers under the current single-device proxy shape: smoke `app_webview_connect`, `set_active_context`, and `multi-tab` against a standard `ios_webkit_debug_proxy -c <UDID>:9322` session.
- [x] Diagnostic: if a redirect ever fails to resolve, the caller observes a shorter-than-expected array rather than an opaque WebSocket error against an undefined URL.

## Out of Scope

- Changes to `isDeviceRedirect`'s regex — the host:port pattern is treated as correct for this fix.
- Retries on redirect failure — the fall-through behaviour remains "swallow and return `[]` for that entry", matching the current semantics.
- Caching of redirect results across `listTargets` calls.
- Changes to callers (`app_webview_connect`, `set_active_context`) — they already assume the returned array contains real targets.

## Related

- PR #633 (recent redirect-follow extension), #592 (bundle-aware target classification), #606 (bundle-aware classifier).

shaun0927:
## Summary

- Closes #648. Replaces the all-or-nothing `parsed.every(isDeviceRedirect)` branch in `WebKitClient.listTargets()` with per-entry resolution so a mixed `/json` response (redirect stubs + inline page targets) no longer leaks stubs with `undefined` `webSocketDebuggerUrl` to callers.
- Each entry is handled independently: redirect stubs are expanded via their inner `/json`, inline targets pass through, and a failure on one redirect only contributes `[]` — the rest of the array still resolves.
- Adds `tests/unit/webkit-client-list-targets.test.ts` with five cases (all-inline / all-redirect / mixed / mixed-with-failure / empty) that exercise the real code path by spying on private `httpGet`, not by stubbing `listTargets` at the interface level.

## Why

`app_webview_connect` and other callers consume `listTargets()` results directly. Under the old guard, a mixed `/json` shape from `ios-webkit-debug-proxy` caused the `else` branch to run and return redirect stubs untouched, producing an opaque WebSocket error with no diagnostic pointing back at the unresolved redirect. See #648 for full evidence, reproduction, and success criteria.

## Test plan

- [x] `npm test -- tests/unit/webkit-client-list-targets.test.ts` — 5/5 pass.
- [x] `npm test -- tests/unit/proxy-timing.test.ts tests/unit/multi-tab.test.ts tests/unit/app-webview-connect.test.ts tests/unit/set-active-context.test.ts` — pass, no regression.
- [x] `npm run build` — exits 0.
- [ ] Post-merge live smoke: `app_webview_connect` against a real `ios-webkit-debug-proxy` session to confirm no behavior change in the standard single-device path.

### DIFF
--- a/src/webkit/client.ts
+++ b/src/webkit/client.ts
@@ -163,25 +163,23 @@ export class WebKitClient extends EventEmitter implements BrowserBackend {
     // ios-webkit-debug-proxy device-list mode: the top-level /json returns
     // redirect entries of the form { url: "host:PORT" } with no
     // webSocketDebuggerUrl. Follow each redirect to get real page targets.
+    // A single /json response may mix redirect stubs with inline page targets,
+    // so resolve per-entry rather than all-or-nothing.
     const isDeviceRedirect = (t: WebKitTarget) =>
       !t.webSocketDebuggerUrl && typeof t.url === 'string' && /^\S+:\d+$/.test(t.url);
 
-    let targets: WebKitTarget[];
-    if (parsed.length > 0 && parsed.every(isDeviceRedirect)) {
-      const followed = await Promise.all(
-        parsed.map(async (t) => {
-          try {
-            const redirectJson = await this.httpGet(`http://${t.url}/json`);
-            return JSON.parse(redirectJson) as WebKitTarget[];
-          } catch {
-            return [] as WebKitTarget[];
-          }
-        }),
-      );
-      targets = followed.flat();
-    } else {
-      targets = parsed;
-    }
+    const expanded = await Promise.all(
+      parsed.map(async (t) => {
+        if (!isDeviceRedirect(t)) return [t];
+        try {
+          const redirectJson = await this.httpGet(`http://${t.url}/json`);
+          return JSON.parse(redirectJson) as WebKitTarget[];
+        } catch {
+          return [] as WebKitTarget[];
+        }
+      }),
+    );
+    const targets = expanded.flat();
 
     // ios-webkit-debug-proxy doesn't include an `id` field — derive from webSocketDebuggerUrl
     for (const t of targets) {
--- a/tests/unit/webkit-client-list-targets.test.ts
+++ b/tests/unit/webkit-client-list-targets.test.ts
@@ -1,0 +1,182 @@
+/**
+ * Tests for WebKitClient.listTargets redirect resolution (Issue #648).
+ *
+ * Covers:
+ * - All-inline input: no redirect HTTP call is issued.
+ * - All-redirect input: every stub is expanded.
+ * - Mixed input: redirect stubs are expanded, inline targets pass through.
+ * - Mixed input with one failing redirect: that entry contributes [], others resolve.
+ * - Empty input: returns [] with no redirect calls.
+ */
+
+import { WebKitClient } from '../../src/webkit/client';
+
+type ClientPrivate = {
+  httpGet: (url: string) => Promise<string>;
+  listTargets: () => Promise<Array<{ id?: string; url?: string; webSocketDebuggerUrl?: string }>>;
+};
+
+function asPrivate(client: WebKitClient): ClientPrivate {
+  return client as unknown as ClientPrivate;
+}
+
+describe('WebKitClient.listTargets redirect resolution', () => {
+  let client: WebKitClient;
+
+  beforeEach(() => {
+    client = new WebKitClient({ host: 'localhost', port: 9221 });
+  });
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  it('passes through inline targets without issuing redirect requests', async () => {
+    const inline = [
+      {
+        id: 'page-1',
+        title: 'Inline A',
+        url: 'https://example.com/a',
+        webSocketDebuggerUrl: 'ws://localhost:9322/devtools/page/page-1',
+      },
+      {
+        id: 'page-2',
+        title: 'Inline B',
+        url: 'https://example.com/b',
+        webSocketDebuggerUrl: 'ws://localhost:9322/devtools/page/page-2',
+      },
+    ];
+
+    const spy = jest
+      .spyOn(asPrivate(client), 'httpGet')
+      .mockResolvedValueOnce(JSON.stringify(inline));
+
+    const targets = await client.listTargets();
+
+    expect(spy).toHaveBeenCalledTimes(1);
+    expect(spy).toHaveBeenCalledWith('http://localhost:9221/json');
+    expect(targets).toHaveLength(2);
+    expect(targets.map((t) => t.id)).toEqual(['page-1', 'page-2']);
+    expect(targets.every((t) => typeof t.webSocketDebuggerUrl === 'string')).toBe(true);
+  });
+
+  it('expands every redirect stub when the top-level /json is all redirects', async () => {
+    const topLevel = [{ url: 'localhost:9322' }, { url: 'localhost:9323' }];
+    const firstDevice = [
+      {
+        id: 'page-a',
+        title: 'Device 1 page',
+        url: 'https://a.example',
+        webSocketDebuggerUrl: 'ws://localhost:9322/devtools/page/page-a',
+      },
+    ];
+    const secondDevice = [
+      {
+        id: 'page-b',
+        title: 'Device 2 page',
+        url: 'https://b.example',
+        webSocketDebuggerUrl: 'ws://localhost:9323/devtools/page/page-b',
+      },
+    ];
+
+    const spy = jest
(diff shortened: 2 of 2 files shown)


shaun0927:
## Merge-readiness review (self-audit)

**Content check**: The per-entry resolution replaces the all-or-nothing `parsed.every(isDeviceRedirect)` branch that #648 identified as leaking stub entries with `undefined webSocketDebuggerUrl` on mixed `/json` responses. The new shape (per-entry redirect expansion, failure on one redirect contributes `[]`, inline passthrough) matches the #648 acceptance criterion. Closes-#648 wording present.

**Test coverage**: New `tests/unit/webkit-client-list-targets.test.ts` (5 cases: all-inline, all-redirect, mixed, mixed-with-failure, empty) exercises the real code path via `httpGet` spy — not the interface-level stub pattern. Existing suites (`proxy-timing`, `multi-tab`, `app-webview-connect`, `set-active-context`) still pass, so no regression on the standard single-device path.

**Bot reviews**: None on file. No reviewer requests outstanding.

**CI**: `lint` ✅, `build` skipped, WebView smoke ✅, `test` ❌ — the failing suite is the same `tests/unit/app-type-element.test.ts` timeout flake affecting every other PR opened today (and `develop` head run 24782234025). The rest of the headless smoke matrix has its own unrelated failures on `develop` too (`Flutter VM-Service`, `Native SimulatorKit HID`). None of these touch the listTargets code path.

**Verdict**: ready to merge as-is. No fix commits needed on this branch.

Issue #641, opened 2026-04-23
[P2] iOS Universal Link channels in Simulator — Notes helper, Messages/Mail limitation, Safari Smart Banner spike, inline log capture (S23 close-gate)


shaun0927:
## Summary

OpenSafari MCP has no end-to-end path for tapping **Universal Links** inside
iOS Simulator first-party apps (Notes / Messages / Mail) or for capturing the
Safari "Open in App" Smart Banner flow. Today the only observable channel is
`xcrun simctl openurl`, which Apple reviewers do not consider equivalent to a
real Universal Link tap; App Review close-gate evidence on the consumer side
(Omofictions/Omofictions-App#26) therefore rests on indirect arguments.

This issue tracks the gap and, after a codebase-level feasibility sweep, breaks
the work into four bounded sub-tasks (T1–T4). Each sub-task carries its own
acceptance criteria, verification steps, and ships as an independent PR.

## Environment

- OpenSafari: `develop` (PRs target `develop`; release merges to `main`)
- iOS Simulator: iPhone 17 Pro / iOS 26.4 (UDID `D7D26213-C3E9-4623-BCCB-984CDF5D0793`); iPhone SE 2nd-gen as secondary repro target
- Host: macOS Apple Silicon
- App under test: Flutter release/QA build (`com.omofictions.omofictionsApp`,
  bundle id team `9S9F688T87`)
- `applinks:omofictions.com` entitlement + apex `/.well-known/apple-app-site-association` 200 + `application/json` confirmed

## Feasibility verdict (codebase sweep)

| # | Request                                  | Verdict                                | Rationale (with pointers)                                                                                                                                                   |
|---|------------------------------------------|----------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| 1 | Notes paste-and-tap helper               | **Feasible**                           | Pasteboard infra is done (`src/tools/pasteboard-input.ts` → `typeViaPasteboard`), accessibility-bridge tap is done (`app_tap_element`). Missing glue tool only.             |
| 2 | Messages / Mail URL-tap simulation       | **Infeasible — document as limitation** | Simulator does not provision iMessage/SMS; no `simctl` verb synthesises inbound messages. Mail.app has no public API for programmatic composition.                          |
[...]

PR #654, opened 2026-04-23
feat(app-notes-paste-and-tap-url): reviewer-equivalent UL tap via Notes (#641 T1)

shaun0927:
## Summary

Adds a new MCP tool `app_notes_paste_and_tap_url` that drives the
**reviewer-equivalent Universal Link channel** — paste the URL into
Notes, let iOS Data Detector produce a tappable link, and press it —
through a single tool call. This covers the close-gate blocker in #641 (sub-task T1).

Apple documentation does not accept `xcrun simctl openurl` as
equivalent to a real Universal Link tap, so consumer teams (e.g.
Omofictions-App#26) have to reconstruct the Notes-paste flow manually.
This tool makes that channel deterministic from the MCP side.

## Implementation

- `src/tools/app-notes-paste-and-tap-url.ts` — glue that:
  1. Launches Notes (`com.apple.mobilenotes`) via `SimulatorManager.launchApp`.
  2. Waits for a focusable editor by polling the AX tree for role in
     `['AXTextArea', 'AXTextView', 'AXTextField']` (configurable `focusTimeoutMs`).
  3. Focuses the editor via `AccessibilityBridge.press`.
  4. Paste-injects the URL via the existing `typeViaPasteboard` from
     `src/tools/pasteboard-input.ts` (already handles the iOS 16+
     paste-permission dialog and restores the original pasteboard).
  5. Polls the AX tree for `AXLink` whose label contains the URL or its
     host (configurable `linkTapTimeoutMs`). If exactly one AXLink is
     present, we take it as a fallback (Data Detector's only link).
  6. Presses the detected link and returns the matched `linkElement` path + label.
- `src/tools/index.ts` — tool registered next to `app_deeplink`.
- `tests/unit/app-notes-paste-and-tap-url.test.ts` — 9 unit tests covering:
  - Happy path (URL literal match)
  - Host-only label match
  - Missing URL
  - Scheme-less URL
  - Editor never appears
  - Data Detector never produces a link
  - Notes launch failure
  - Press-detected-link fails (`PRESS_NOT_ACTIONABLE`)
  - Tool registration

No changes to the existing `app_open_url` / `app_deeplink` tools — this
PR is purely additive.

## Verification

- [x] `npx tsc --noEmit` clean
- [x] `npx jest tests/unit/app-notes-paste-and-tap-url.test.ts` → 9/9 pass
- [ ] Manual repro (iPhone 17 Pro, iOS 26.4, Flutter app
      `com.omofictions.omofictionsApp`): invoke tool with
      `https://omofictions.com/detail/abc`, confirm app foregrounds at
      the expected route. *(pending — will be attached to #641 before merge)*

## Scope

This is T1 of four sub-tasks in #641:

- **T1 (this PR)** — `app_notes_paste_and_tap_url` tool
- T2 — Doc: UL channels + simulator limitations (separate PR)
- T3 — Safari Smart Banner spike (separate follow-up issue)
- T4 — `captureLogs` option on `app_open_url` / `app_deeplink` (separate PR)

Ref: #641

## Test plan

- [ ] CI green (build + unit tests)
- [ ] Manual repro on iOS 26.4 Simulator with `applinks:`-entitled Flutter app
- [ ] Fallback: confirm `isError: true` when Notes is force-uninstalled

### DIFF
--- a/src/tools/app-notes-paste-and-tap-url.ts
+++ b/src/tools/app-notes-paste-and-tap-url.ts
@@ -211,17 +211,30 @@ async function waitForDetectedLink(
         { role: 'AXLink' },
         { deviceId, maxResults: 20 },
       );
+      // Two-pass scan: prefer an exact URL-literal match over a host-only match.
+      // A user note may contain several links to the same host (for example
+      // `https://example.com` above the requested `https://example.com/detail/abc`);
+      // tapping the first host-containing label would route to the wrong path
+      // and still report success, which breaks deterministic deep-link
+      // verification.
       for (const match of res.matches) {
         const label = (match.label ?? '').toLowerCase();
-        if (label.includes(lowerUrl) || (host && label.includes(host))) {
+        if (label.includes(lowerUrl)) {
           return { path: match.path, label: match.label ?? url };
         }
       }
-      // Fallback: if exactly one AXLink is present, assume it's the one we pasted.
-      if (res.matches.length === 1) {
-        const sole = res.matches[0];
-        return { path: sole.path, label: sole.label ?? url };
+      if (host) {
+        for (const match of res.matches) {
+          const label = (match.label ?? '').toLowerCase();
+          if (label.includes(host)) {
+            return { path: match.path, label: match.label ?? url };
+          }
+        }
       }
+      // No single-link fallback: Notes is launched into prior state, so an
+      // unrelated pre-existing AXLink could satisfy a blanket `matches.length
+      // === 1` heuristic and yield a false-positive tap. Callers that need a
+      // looser match should widen the input URL (or host) themselves.
     } catch {
       /* tree may be mid-transition — retry */
     }
--- a/src/tools/app-notes-paste-and-tap-url.ts
+++ b/src/tools/app-notes-paste-and-tap-url.ts
@@ -1,0 +1,250 @@
+/**
+ * app_notes_paste_and_tap_url — reviewer-equivalent Universal Link tap through Notes.app.
+ *
+ * The iOS Simulator only exposes `xcrun simctl openurl` as a URL-opening
+ * channel, which Apple documentation does not treat as equivalent to a real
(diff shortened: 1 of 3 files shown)


shaun0927:
## Ready to merge — second-pass review complete

Re-reviewed this PR end-to-end against both reviewers (human + Codex) and addressed every actionable item. Change set now stands as:

### Codex P1 issues — fixed in `6530cd61`
- **Prefer exact URL match over host-based fallback** (`src/tools/app-notes-paste-and-tap-url.ts`): `waitForDetectedLink` now scans every `AXLink` twice — first pass seeks an exact URL-literal match, second pass falls back to host match. Previously the first label containing *either* the full URL or just the host returned, so a stray same-domain link (e.g. `https://example.com` above the pasted `https://example.com/detail/abc`) could win and silently route to the wrong path.
- **Remove unconditional single-link fallback tap**: the `matches.length === 1` branch is gone. Notes is launched into prior state so any pre-existing unrelated link could satisfy that heuristic and yield a false-positive tap.

### Test additions
- `prefers exact URL match over a host-only match earlier in the list` — freezes the new two-pass ordering.
- `does not tap a sole unrelated AXLink as a fallback` — guards against the old single-link-fallback regression.
- Existing `matches link by URL host when label only shows the host` still passes (host-only fallback still wins when no full-URL match exists, which is the intended behaviour).

### Local verification
- `npx jest tests/unit/app-notes-paste-and-tap-url.test.ts` → **11/11 pass**
- `npx tsc --noEmit` → clean

### CI note
The `test` job may continue to fail on `tests/unit/app-type-element.test.ts` (`Exceeded timeout of 5000 ms`). This is a pre-existing timing flake that also fails on the current `develop` tip ([run 24782234025](https://github.com/shaun0927/opensafari/actions/runs/24782234025): `13 failed, 1907 passed`, same test file). This PR does not touch `app-type-element.test.ts`, so the failure is not a regression introduced here.

Ready for merge once a reviewer confirms the tightened link-selection contract.

Issue #640, opened 2026-04-23
[P2] Native iOS Simulator network toggle for Flutter offline verification (device_network_set/get)


shaun0927:
# [P2] Native iOS Simulator network toggle for Flutter offline verification

## Context

OmoFictions-App (Flutter, iOS Simulator) needs an automated check that a purchased chapter can be opened and fully read **from a cold start in airplane mode** (close-gate scenario S27). Today this is only verified by a human manually toggling airplane mode (`Features → Toggle Airplane Mode`, `⌥⌘A`) because opensafari has no way to cut external network traffic for a native Flutter app running in the simulator.

- Upstream blocker: https://github.com/Omofictions/Omofictions-App/issues/31 (S27 — "offline reading, close-gate scenario")
- App flow under test: `ViewerPage._loadContent()` reads a local pre-parsed cache first and falls back to the network. Any offline-cache regression is masked in CI because the API fallback silently succeeds.

## Why existing tools do not solve this

| Tool | Scope today | Gap |
|---|---|---|
| `mcp__opensafari__network_offline` | Injects a JS `window.fetch` / `XMLHttpRequest.prototype.open` override through `networkInterceptor.setOffline()` (`src/tools/network-intercept.ts:17`, `src/network-interceptor.ts:92`). WebKit page context only. | Does **not** affect `URLSession` or `dart:io HttpClient`. |
| `mcp__opensafari__network_throttle` | Same JS-injection approach — adds latency to `fetch` / `XHR` in the page. | Same gap. Not a native-traffic control. |
| `mcp__opensafari__flutter_network` | Starts a local Node proxy and writes `Apple Global Domain` proxy defaults via `xcrun simctl spawn defaults write`. Observer-oriented (HAR capture). | Apps that bypass proxy settings (including native `URLSession` configured without proxy) still reach the network. Not a blocking primitive. |

> **Issue body correction vs. original draft:** the original draft said `network_offline` uses `Network.setEmulateOfflineState` via the WebKit Inspector Protocol. It actually uses JS fetch/XHR shimming — the failure mode is the same (WebKit-only) but future PR descriptions should cite the correct implementation.

## Why the originally proposed `simctl status_bar` approach does NOT work

The first draft suggested:

```
xcrun simctl status_bar <udid> override --dataNetwork none --wifiMode searching
```

[...]

PR #656, opened 2026-04-23
feat(device-network): scaffold device_network_set/get tools (#640 PR 1/6)

shaun0927:
## Summary

- Adds `device_network_set` and `device_network_get` MCP tools — scaffold only. Real backends (`pfctl`, NLC) land in follow-up PRs 2–6 tracked in #640.
- Documents why existing `network_offline` / `network_throttle` / `flutter_network` cannot deliver a native `SocketException` on Flutter (WebKit-only JS shims / HTTP-proxy observer, respectively).
- Zero behavior change for existing tools; all 114 unit-test suites (1693 tests) continue to pass.

## Why this PR is valuable on its own

Shipping the schema + registration ahead of the mechanism unblocks downstream tasks:
- Callers can start wiring UI / scenario scripts against the stable API surface.
- Any request for `mode: "offline" | "airplane"` currently returns a structured `not_implemented` error that names the tracking issue (#640) and the PR sequence — this is intentional and far safer than silently returning `simctl status_bar override --dataNetwork none`, which is cosmetic only.

## Background (see issue #640 for full context)

The Omofictions S27 close-gate test needs a purchased chapter to open and read fully after the simulator is taken offline. Today that scenario can only be verified by a human toggling airplane mode (`⌥⌘A`), because:

| Existing tool | Implementation | Why it can't gate S27 |
|---|---|---|
| `network_offline` | JS `window.fetch` / `XMLHttpRequest.prototype.open` shim (`src/network-interceptor.ts:92`) | WebKit page scope only. Does not touch `URLSession` / `dart:io HttpClient`. |
| `network_throttle` | Same JS-shim approach with latency | Same scope gap. |
| `flutter_network` | Node HTTP proxy + `xcrun simctl spawn defaults write ... WebKitProxyDefaultsKey` | Observer-oriented; apps that bypass proxy settings still reach the network. |

The iOS Simulator shares the host network stack, so the real fix is a host-level block (`pfctl`, NLC). Issue #640 enumerates the mechanism options and their tradeoffs; this PR only lays down the API surface.

## Changes

- **`src/tools/device-network.ts`** (new): handler + schema for both tools. In-memory state map keyed by device UDID. Device resolution mirrors `device-shutdown.ts` (`udid` param → `getSessionManager().getSoleDeviceId()` → `SimulatorManager.listBooted()[0]`). `mode: "online"` is idempotent and records state; `mode: "offline" | "airplane"` returns a structured `not_implemented` error with a link back to #640.
- **`src/tools/index.ts`**: register the two tools alongside the existing WebKit-scoped network tools, with a comment pointing at #640.
- **`tests/unit/device-network.test.ts`** (new, 12 cases): schema shape, mode/mechanism validation, online idempotence, UDID resolution (explicit vs sole vs booted fallback), `device_network_get` default + after-set behavior.

## Test plan

- [x] `npm run build` — clean.
- [x] `npx tsc -p tsconfig.json --noEmit` — no type errors.
- [x] `npx jest tests/unit/device-network.test.ts` — 12 passed, 12 total.
- [x] `npx jest tests/unit/` — 114 suites / 1693 tests passed (no regressions).
- [ ] CI green on this PR.

## Follow-up PRs (tracked in #640)

- **PR 2** — mechanism abstraction (`NetworkBlocker` interface + `PfctlBlocker` / `NlcBlocker` / `AutoBlocker` stubs with mocked exec).
- **PR 3** — real `pfctl` mechanism + crash-safe cleanup + sudoers docs.
- **PR 4** — Network Link Conditioner fallback.
- **PR 5** — `docs/tools/device-network.md` + Omofictions#31 S27 example.
- **PR 6** *(optional)* — reusable Flutter fixture app under `tests/fixtures/` exposing a `probeNetwork()` service extension.

Each lands independently so that mechanism research on PR 3/4 does not block the API surface.

Closes: part of #640 (scaffold). Full close on PR 5.

### DIFF
--- a/src/tools/device-network.ts
+++ b/src/tools/device-network.ts
@@ -1,0 +1,171 @@
+import { MCPServer } from '../mcp-server';
+import { SimulatorManager } from '../simulator';
+import { getSessionManager } from '../session-manager';
+
+export type DeviceNetworkMode = 'online' | 'offline' | 'airplane';
+export type DeviceNetworkMechanism = 'pfctl' | 'nlc' | 'auto';
+export type DeviceNetworkResolvedMechanism = 'pfctl' | 'nlc' | null;
+
+export interface DeviceNetworkStateEntry {
+  mode: DeviceNetworkMode;
+  mechanism: DeviceNetworkResolvedMechanism;
+  activeSince: string | null;
+}
+
+const DEFAULT_STATE: DeviceNetworkStateEntry = {
+  mode: 'online',
+  mechanism: null,
+  activeSince: null,
+};
+
+const stateByDevice = new Map<string, DeviceNetworkStateEntry>();
+
+export function __resetDeviceNetworkStateForTests(): void {
+  stateByDevice.clear();
+}
+
+export function getDeviceNetworkState(deviceId: string): DeviceNetworkStateEntry {
+  return stateByDevice.get(deviceId) ?? { ...DEFAULT_STATE };
+}
+
+export function setDeviceNetworkState(deviceId: string, entry: DeviceNetworkStateEntry): void {
+  stateByDevice.set(deviceId, entry);
+}
+
+async function resolveDeviceId(explicit: string | undefined): Promise<string | null> {
+  if (explicit) return explicit;
+  const sole = getSessionManager().getSoleDeviceId();
+  if (sole) return sole;
+  const manager = new SimulatorManager();
+  const booted = await manager.listBooted();
+  return booted[0]?.udid ?? null;
+}
+
+function jsonResponse(body: unknown, isError = false) {
+  return {
+    content: [{ type: 'text' as const, text: JSON.stringify(body) }],
+    ...(isError ? { isError: true as const } : {}),
+  };
+}
+
+export function registerDeviceNetworkSetTool(server: MCPServer): void {
+  server.registerTool(
+    {
+      name: 'device_network_set',
+      description:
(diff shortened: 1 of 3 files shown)


shaun0927:
## Ready to merge — second-pass review complete

Re-reviewed end-to-end. Codex's P2 finding is addressed in `26e8ac9a`. The scaffold contract is unchanged (`offline`/`airplane` still return `not_implemented` pointing at #640), but the device-resolution policy is tightened so `device_network_set/get` cannot silently act on the wrong simulator.

### Codex P2 issue — fixed in `26e8ac9a`
**Validate UDID instead of defaulting to arbitrary booted device** (`src/tools/device-network.ts`).

New `resolveDeviceId` policy:

- **Explicit `udid`**: validated against `listBooted()`. Returns a structured `udid_not_booted` error (with `requestedUdid` + `bootedUdids`) instead of silently accepting a typo/shutdown UDID and persisting state under that key.
- **Omitted `udid`**:
  1. Use the session's sole-device binding if present (authoritative — we don't second-guess against `listBooted` here to avoid racing a just-shutdown device mid-session).
  2. Fall back to `listBooted()` only when exactly one simulator is booted.
  3. Multiple booted simulators → `ambiguous_device` error (refuses to pick `booted[0]`).
  4. No booted simulators → `no_booted_device`.

### Test additions
- `rejects explicit udid when it is not currently booted (typo / shutdown guard)`
- `rejects defaulting to an arbitrary simulator when multiple are booted (ambiguous_device)`
- `reports no_booted_device when nothing is booted and no udid is passed`
- `device_network_get rejects explicit udid that is not booted`
- `honors explicit udid when the device is booted` (reworks the prior `honors explicit udid over booted lookup`)

### Local verification
- `npx jest tests/unit/device-network.test.ts` → **16/16 pass**
- `npx tsc --noEmit` → clean

### CI note
The `test` job may continue to fail on `tests/unit/app-type-element.test.ts` timeouts — pre-existing flake on `develop` ([run 24782234025](https://github.com/shaun0927/opensafari/actions/runs/24782234025)), not touched by this PR.

Ready for merge.

Issue #651, opened 2026-04-23
[bug] app_handle_alert AX-bridge fails on Korean-locale push permission dialog; screenshot tool misreads simctl stderr


shaun0927:
## TL;DR

On Korean-locale iOS 26.4 simulators, OpenSafari's tooling cannot close the standard iOS push-notification permission dialog via `app_handle_alert`/`app_alert_handle` because the native AX-bridge dump returns an empty/failed tree for the `UNUserNotificationCenter` permission sheet, and no lower-level keyboard fallback exists. Separately, `app_screenshot_native` treats transient `simctl io screenshot` surface-timeouts as hard failures instead of retrying them, blocking scripted screenshot sweeps after rapid swipe/scroll sequences.

Two orthogonal reliability bugs, both reproducible on `iPhone 17 Pro` sim running `iOS 26.4` under `ko-KR`:

- **Bug 1** — `app_handle_alert` / `app_alert_handle` cannot dismiss the standard push-notification permission dialog (AX Tier 1 sees zero buttons, AppleScript Tier 2 misses, `ax-bridge-native dump` itself fails). Workaround today: hand-supplying labels to `app_alert_handle`, which then also fails because the bridge dump fails.
- **Bug 2** — `app_screenshot_native` raises an error when `simctl io screenshot` emits `Note: No display specified.` on stderr or returns transient `Timeout waiting for screen surfaces` after rapid input events. Same command succeeds on manual retry ~4–5 s later.

## Environment

- macOS host, Xcode iOS Simulator
- Device: iPhone 17 Pro (UDID `D7D26213-C3E9-4623-BCCB-984CDF5D0793`), runtime `iOS 26.4 (com.apple.CoreSimulator.SimRuntime.iOS-26-4)`
- Locale: `ko-KR` (system / keyboard)
- App under test: Flutter `com.omofictions.omofictionsApp` (debug build)
- OpenSafari dist: `/Users/jh0927/opensafari/dist/ax-bridge-native`

### Bug 1 — Korean-locale push permission dialog

1. Launch a Korean-locale simulator and set `xcrun simctl ui <udid> appearance dark` + `xcrun simctl ui <udid> content_size accessibility-extra-extra-large`.
2. `mcp__opensafari__app_launch { bundleId: "com.omofictions.omofictionsApp" }`.
3. iOS OS presents the standard push-notification permission dialog:

   > ''Omofictions App'에서 알림을 보내고자 합니다.
[...]

PR #659, opened 2026-04-23
feat(app-handle-alert): Tier 2.5 keyboard fallback for system dialogs (#651)

shaun0927:
## Summary
- Add a Tier 2.5 keyboard fallback to `app_handle_alert` that fires after AX-scan (Tier 1) and AppleScript label match (Tier 2) both miss.
- `action: \"accept\"` sends `Return` (key code 36); `action: \"dismiss\"` sends `Escape` (key code 53).
- Gated via new optional param `keyboardFallback: boolean` (default `true`); strategy is reported in the response as `strategy: 'keyboard-fallback'`.

Addresses bug #1 from #651 — ko-KR iOS 26.4 push-permission dialogs where ax-bridge-native returns an empty tree.

## Test plan
- [x] `npx jest tests/unit/app-handle-alert-keyboard-fallback.test.ts` — 8/8 pass
- [x] `npm run build` — green
- [ ] Manual: ko-KR Flutter cold launch, call `app_handle_alert { action: \"dismiss\" }` — expect `dismissed: true, strategy: 'keyboard-fallback'`
- [ ] Manual: same dialog, call `app_handle_alert { action: \"accept\" }` — expect `dismissed: true, strategy: 'keyboard-fallback'`

## Notes for reviewer
- Keyboard fallback is intentionally last-resort: it only fires when the higher tiers have missed. AX-scan success short-circuits and never reaches the keyboard path.
- The escalation is safe for system dialogs (where accept/dismiss is unambiguous), but callers who need strict label verification can opt out via `keyboardFallback: false`.
- A follow-up issue will track the Swift-level fix for the ax-bridge-native dump failure that necessitates this fallback in the first place. Once that lands, Tier 1 should again be the primary path — this commit does not remove or regress it.

### DIFF
--- a/src/tools/app-handle-alert.ts
+++ b/src/tools/app-handle-alert.ts
@@ -26,7 +26,12 @@ const DEFAULT_DEVICE_WIDTH = 393;
 const AX_POLL_TIMEOUT_MS = 3000;
 const AX_POLL_INTERVAL_MS = 250;
 
-type Strategy = 'ax-scan' | 'applescript-sheet' | 'permission-reset' | 'none';
+type Strategy =
+  | 'ax-scan'
+  | 'applescript-sheet'
+  | 'keyboard-fallback'
+  | 'permission-reset'
+  | 'none';
 type Surface = 'simulator_chrome' | 'system_dialog_unknown' | 'app_content';
 type FallbackMode = 'permission_reset' | 'none';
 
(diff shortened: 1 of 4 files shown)


shaun0927:
## Merge-readiness recap (autopilot sweep @ `6cc56d16`)

### Scope check ✅
PR adds Tier 2.5 keyboard fallback. Implementation is gated on the new `keyboardFallback` param (default `true`) and only fires when Tier 1 (AX-scan) and Tier 2 (AppleScript label match) both miss. Short-circuits with `strategy: 'keyboard-fallback'` when the keystroke lands; otherwise falls through to diagnostics or Tier 3 (`permission_reset`) unchanged.

### Bot review ✅
Codex reviewed `b2c4d521b8` and posted the standard meta comment only — no suggestions or blockers surfaced.

### CI follow-up
- `b2c4d521`: test job was red because two legacy suites (`app-handle-alert.test.ts`, `app-handle-alert-permission-reset.test.ts`) asserted exact `strategy_attempted` arrays like `['ax-scan', 'applescript-sheet']` that now include `'keyboard-fallback'`.
- `6cc56d16` (this push): those three tests now pass `keyboardFallback: false` so they exercise the tier they were written for. The dedicated Tier 2.5 coverage in `tests/unit/app-handle-alert-keyboard-fallback.test.ts` stays as-is (8/8 pass).
- Latest run (`24825525162`): lint ✅, test ❌ — but the only remaining failures are the `tests/unit/app-type-element.test.ts` timeout flakes that reproduce on `develop@b2445dc3` itself and are **not introduced by this PR**. See develop run `b2445dc3` → same 14 timeouts across `types into element found by label`, `Tier 1.5 AX press focus`, `Tier 3 readback verification`, etc.

### Verdict
Merge-ready once the pre-existing develop flake is unblocked (orthogonal issue). This PR's own behavior and tests are green.

Issue #639, opened 2026-04-23
iOS Simulator host→app text input lane is non-deterministic (IME reinterpretation, segmented-field drops, silent paste, AX recovery) — blocks Flutter signup/auth verification


shaun0927:
## Summary

In an OpenSafari-only verification lane against a Flutter iOS app on iOS 26 Simulator, four distinct **host → simulator text input** failures make signup/auth flows impossible to verify deterministically. Each is reproducible on a clean simulator with a QA build of the app. Filed as a single consolidated report because the four problems share infrastructure (text-injection backend, alert handling, AX recovery), but they are intended to land as **four independent PRs** that can be merged in priority order.

Codebase verification (2026-04-23) confirmed all four are current, real issues with no existing remediation:

- Problem 1 (IME): `src/tools/sim-hid-input-backend.ts:214` sends raw HID assuming US-ABC; `src/tools/keyboard-layout.ts` can detect non-Latin layouts but cannot switch them. The pasteboard backend (`src/tools/pasteboard-input.ts`) is the only existing escape hatch.
- Problem 2 (OTP): `sim-hid-input-backend.ts:223-244` sends per-character HID with no inter-key delay parameter; `app-type-element.ts:273-281` has no segmented-field awareness.
- Problem 3 (paste): `pasteboard-input.ts:139-157` treats absent permission dialog as success; no readback verification of field value after Cmd+V.
- Problem 4 (AX recovery + alerts): `app-handle-alert.ts:298-325` only dismisses the first candidate (no stacked-alert loop); `pollForDismissal()` does not refresh AX after dismissal; `app-tree.ts` has no automatic retry on `DEVICE_CONTENT_ROOT_EMPTY`. The Korean label corpus is present in `app-handle-alert-labels.ts` but iOS 26-specific button text (e.g. Apple Intelligence onboarding) is not classified.

## Environment

- OpenSafari: current `develop` HEAD
- iOS Simulator: iPhone 17 Pro / iOS 26.4 (UDID `D7D26213-C3E9-4623-BCCB-984CDF5D0793`) and iPhone SE (2nd gen)
- Host: macOS, Apple Silicon
- App under test: Flutter 3.x release/QA build (Flutter release build on Xcode 26+)
- Input method on host: Korean 2-Set (default system IME at the time of capture — likely load-bearing for Problem 1)

---

### Repro
1. Boot iPhone 17 Pro / iOS 26.4 with Korean 2-Set as the active host IME.
[...]

PR #669, opened 2026-04-23
fix(app-tree): retry on DEVICE_CONTENT_ROOT_EMPTY after alert dismissal (#639 Problem 4a)

shaun0927:
## Summary
Adds an automatic retry loop with backoff when `app_tree` hits `DEVICE_CONTENT_ROOT_EMPTY` — the transient AX state iOS leaves immediately after a system alert is dismissed. Closes the most-impactful slice of #639 Problem 4.

## Why
Issue #639 Problem 4a: callers running `app_handle_alert` followed by `app_tree` get a hard `DEVICE_CONTENT_ROOT_EMPTY` for a few seconds while iOS repopulates the foreground app's AX root. The previous behaviour required either defensive caller-side sleeps or a full simulator relaunch — both make Flutter signup/auth lanes brittle.

## Changes
- `src/tools/app-tree.ts`
  - New exported `dumpTreeWithRetry(bridge, options, maxRetries, sleep)` — retries `bridge.dumpTree()` with backoff `[250, 500, 1000]` ms when the underlying error is `AccessibilityBridgeError` with code `DEVICE_CONTENT_ROOT_EMPTY`. Other errors are re-thrown immediately.
  - The no-bundleId path (the path that drives raw AX dumps right after permission alerts) now uses the retry wrapper. The bundleId / target-app-context path is left unchanged because `ensureTargetAppContext` already has its own retry/heuristics.
- `tests/unit/app-tree-retry.test.ts`
  - 5 new tests: success-no-retry, retry-then-success, exhaust-then-throw, immediate re-throw on non-empty-root bridge errors, immediate re-throw on plain errors. Backoff schedule asserted.

## Tests
- `npx tsc --noEmit -p tsconfig.json` — clean.
- `npx jest tests/unit/app-tree-retry.test.ts` — 5/5 pass.

## Acceptance (from #639 Problem 4)
- [x] **4a** — `app_tree` retries automatically on `DEVICE_CONTENT_ROOT_EMPTY` (3 retries, backoff `[250, 500, 1000]` ms).
- [ ] **4a** — `app_handle_alert` reports `axRecovered: boolean`. Deferred — the `app_tree` retry alone resolves the user-facing symptom; structuring the recovery signal inside `app_handle_alert` belongs with the next slice.
- [ ] **4b** — `dismissAllVisible` loop for stacked alerts (Apple Intelligence + push permission). Deferred to follow-up.
- [ ] **4c** — Korean push-permission button classification regression fixture. Deferred to follow-up. The label corpus already contains `허용` / `허용 안 함`; the failure observed on iPhone 17 Pro / iOS 26.4 needs a captured AX dump to fixture against, which is out of scope for this PR.

## Backward compatibility
Purely additive. Default retry count and backoff are conservative; the worst case adds ~1.75 s of latency only when the underlying call would otherwise have hard-failed. Successful calls take the existing fast path.

Refs #639 (Problem 4a). Follow-up PR will cover 4b/4c.

### DIFF
--- a/src/tools/app-tree.ts
+++ b/src/tools/app-tree.ts
@@ -5,6 +5,49 @@ import {
   createContextMismatchError,
   ensureTargetAppContext,
 } from './native-app-context';
+import { AccessibilityBridgeError } from '../native/accessibility-bridge';
+import { DEVICE_CONTENT_ROOT_EMPTY } from '../native/ax-bridge-content-root';
+
+const APP_TREE_RETRY_COUNT = 3;
+const APP_TREE_RETRY_BACKOFF_MS = [250, 500, 1000] as const;
+
+/**
+ * Calls `bridge.dumpTree(options)` and retries with backoff on
+ * `DEVICE_CONTENT_ROOT_EMPTY`. The empty-root condition is observed
+ * transiently after a system alert is dismissed (issue #639 Problem 4a) —
+ * iOS takes a beat to repopulate the AX root for the foreground app, and
+ * before that beat the bridge sees zero app-semantics nodes and reports
+ * the structured empty-root error. Re-querying after a short backoff
+ * almost always succeeds; only persistent empty-root after `maxRetries`
+ * surfaces as an error.
+ *
+ * Other bridge errors are re-thrown immediately without retry.
+ *
+ * Exported so unit tests can assert the retry contract without driving
+ * a live simulator.
+ */
+export async function dumpTreeWithRetry(
+  bridge: { dumpTree: (opts: { deviceId?: string; maxDepth?: number }) => Promise<unknown> },
+  options: { deviceId?: string; maxDepth?: number },
+  maxRetries = APP_TREE_RETRY_COUNT,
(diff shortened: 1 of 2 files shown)


shaun0927:
## Self-review — merge status

**Content verdict:** merge-ready.
- Locally: `npx jest tests/unit/app-type-element.test.ts tests/unit/app-tree-retry.test.ts --testTimeout=30000` → **25/25 pass in ~7.9 s**.
- This PR only touches `src/tools/app-tree.ts` (new `dumpTreeWithRetry`) and adds a new test file; it does not touch `app-type-element.ts` or its tests.

**CI test failure is the pre-existing `develop` flake on `tests/unit/app-type-element.test.ts`.**
- Same symptom on `develop@b2445dc3` run [24782234025](https://github.com/shaun0927/opensafari/actions/runs/24782234025) before this PR existed.
- Root cause: the test file has no explicit `jest.setTimeout`, so it uses Jest's 5000 ms default; the GitHub runner occasionally exceeds that. Unrelated to this PR.

No code changes needed on this PR.

PR #671, opened 2026-04-23
feat(app-type-element): attach TEXT_INPUT_LAYOUT_MISMATCH on non-Latin layout (#639 Problem 1)

shaun0927:
## Summary
Attaches a structured `TEXT_INPUT_LAYOUT_MISMATCH` error payload to the existing readback-mismatch response when the simulator's active keyboard layout is confirmed non-Latin. Lets MCP clients / agents programmatically pick the recommended remediation (`backend: "pasteboard"`) instead of inferring intent from a free-form `verify_reason` string.

## Why
Issue #639 Problem 1: with a Korean 2-Set host IME, `app_type_element` sending the ASCII string `qa.signup@example.com` lands as `ㄴ뎁ㅁ.냐후ㅕㅔ@ㄷㅌ므ㅔㅣㄷ.채ㅡ` in the destination field. Even a Swift `CGEvent.keyboardSetUnicodeString` helper produces the same jamo output, which means the failure is on the simulator side (active IM extension or default keyboard layout). There is currently no documented OpenSafari knob to force the simulator's input source to ASCII/US.

Until a native input-source switcher lands, this PR makes the failure **machine-readable** so callers can route around it without parsing prose.

## Changes
- `src/tools/keyboard-layout.ts`
  - New `mismatchHint(expected, actual, detectedLayout)` pure helper. Returns `null` for Latin (QWERTY) layouts and for failed detection. Returns `{ code: "TEXT_INPUT_LAYOUT_MISMATCH", expected, actual, suggestedBackend: "pasteboard", detectedLayout }` for confirmed non-Latin layouts.
  - Module-header limitation note + workaround pointer.
- `src/tools/app-type-element.ts`
  - Module-header doc explaining the limitation and the `backend: "pasteboard"` workaround.
  - `VerifyResult` now carries the raw `observed` AX value alongside the existing `verified` / `verify_method` / `verify_reason` fields, so the caller does not have to re-query the bridge.
  - When the existing mismatch path fires (`verify.verified === false`) AND `mismatchHint(…)` returns a non-null hint (i.e. layout is confirmed non-Latin), the response body now also includes `responseBody.error = hint` in addition to the existing `isError: true` flag.
- `tests/unit/keyboard-layout-mismatch.test.ts`
  - 5 tests: null/Latin no-op, Korean and Japanese hint shape, and JSON-serialisability through MCP.

## Coordination with PR A (#639 Problem 2)
PR A introduces `perKeyDelayMs` for segmented OTP fields — the failure mode there is **dropped characters on a Latin layout**, which gets a different code (`TEXT_INPUT_DROPPED`). The two error codes are mutually exclusive by layout detection in this PR — `mismatchHint()` deliberately returns `null` for Latin layouts so the dropped-character case does not collide. PR A and this PR can land in either order.

## Tests
- `npx tsc --noEmit -p tsconfig.json` — clean.
- `npx jest tests/unit/keyboard-layout-mismatch.test.ts` — 5/5 pass.
- `npx jest tests/unit/app-type-element.test.ts` — 20/20 still pass (no regressions from the `VerifyResult.observed` field addition).

## Acceptance (from #639 Problem 1)
- [x] Structured `TEXT_INPUT_LAYOUT_MISMATCH` payload surfaced when readback diverges AND layout is non-Latin.
- [x] `suggestedBackend: "pasteboard"` field in the payload.
- [x] `detectedLayout` field carries the raw `AppleKeyboards` entry for triage.
- [x] Tool docstring lists the limitation and the `backend: "pasteboard"` workaround.
- [x] Backward-compatible: existing `verified`/`verify_reason`/`isError` surface is unchanged.
- [ ] Native simulator input-source switcher (Swift/IOKit `TIS`). **Out of scope** — tracked as a follow-up if the fail-loud path proves insufficient in practice.

Refs #639 (Problem 1).

### DIFF
--- a/src/tools/app-type-element.ts
+++ b/src/tools/app-type-element.ts
@@ -9,6 +9,29 @@
  * once focused).
  *
  * Works with any app including Flutter — no WebKit/DOM required.
+ *
+ * ## Keyboard-layout limitation (issue #639 Problem 1)
+ *
+ * When `backend: "auto"` (default), text is sent via raw HID keycodes
+ * that assume a US-ABC (QWERTY) software keyboard. If the simulator's
+ * active input source is non-Latin (Korean 2-Set, Japanese Kana, Chinese
+ * Pinyin, …) those keycodes are silently re-composed by the iOS IME into
+ * the corresponding script — producing garbage in the text field.
+ *
+ * There is currently no documented way to switch the simulator's input
+ * source programmatically from the host. Until a native switcher lands,
+ * use `backend: "pasteboard"` which round-trips via the simulator
+ * clipboard (Cmd+V) and is fully keyboard-layout-independent.
+ *
+ * When post-typing readback (`verifyAfterTyping: true`, or the legacy
+ * `verify: true` default) detects divergence AND the detected keyboard
+ * layout is non-Latin, the tool returns `isError: true` with:
+ *
+ *   { code: "TEXT_INPUT_LAYOUT_MISMATCH", expected, actual,
+ *     suggestedBackend: "pasteboard", detectedLayout }
+ *
+ * For Latin layouts where characters are dropped (different failure
+ * mode), see error code `TEXT_INPUT_DROPPED` (PR A / issue #639).
  */
 
 import { execFile } from 'child_process';
@@ -20,6 +43,7 @@ import type { AXNode, AXQuery } from '../native';
 import { resolveDeviceId, getInputBackend, runInputOp } from './native-input-utils';
 import { tryPress } from './app-tap-element';
 import { typeViaPasteboard } from './pasteboard-input';
+import { mismatchHint } from './keyboard-layout';
 
 const execFileAsync = promisify(execFile);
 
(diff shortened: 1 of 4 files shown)


shaun0927:
## Triage update — merge-ready (pending CI)

Reviewed the Codex automated feedback and pushed a follow-up commit that addresses both findings:

- **P1 — PII in `TEXT_INPUT_LAYOUT_MISMATCH` payload**
  Truncation now lives inside `mismatchHint()` itself (24-char echo cap, matches the existing `verify_reason` ceiling). The hint can no longer leak full user text (emails / tokens / passwords) to MCP error logs, and a new `truncated: boolean` flag makes the echo's advisory nature explicit to downstream consumers. Test `truncates long expected/actual values so hint payload never leaks full user text` locks this in.

- **P2 — `detectedLayout` is not guaranteed active**
  Added a `layoutSource: "apple_keyboards_first_entry"` field so callers can see the detection is heuristic (per the module header note — iOS 26.4 exposes no deterministic active-layout signal from the host). `detectedLayout` is still surfaced for triage, but is now explicitly flagged as probabilistic rather than verified-active.

Local: `npx jest tests/unit/keyboard-layout-mismatch.test.ts tests/unit/app-type-element.test.ts` — 26/26 pass. The CI `test` job failure on the prior commit was an `app-type-element.test.ts` Jest 5000 ms timeout flake that reproduces on develop base too (the same tests pass in ~4 s on a clean macOS clone); CI has been rerun.

Happy to split the redaction cap into a named constant + wider module-level refactor if you'd prefer.

Issue #660, opened 2026-04-23
feat(ax-bridge): enumerate UNUserNotificationCenter permission sheet (#651 follow-up)


shaun0927:
## Summary

Tracking issue for the Swift-level portion of #651 that couldn't be landed as part of the initial PR pair (PR A: screenshot retry, PR B: keyboard fallback). The TypeScript-side workarounds ship without this, but full end-to-end reliability for system push-permission dialogs requires an ax-bridge-native fix.

## Context

From #651: on a `ko-KR` iOS 26.4 simulator running a Flutter app, `ax-bridge-native dump` fails to produce a usable tree for the standard `UNUserNotificationCenter` permission sheet rendered by SpringBoard. Observed: exit non-zero with a truncated stderr tail, so the wrapper in `src/native/accessibility-bridge.ts` rejects with `BRIDGE_EXEC_FAILED`. The Tier 2.5 keyboard fallback (PR B) recovers the *dismiss/accept* action, but the diagnostics block (`visibleButtons`, `visibleStaticTexts`, `suggestedLabelsToAdd`) is still empty because the AX dump never returned.

## Root-cause hypothesis

The Swift bridge likely walks the `AXUIElement` tree starting from the *frontmost* application, which is the Flutter app — not SpringBoard — even though SpringBoard is the process actually hosting the permission sheet. Permission sheets render as child windows of SpringBoard's process, not the app's.

## Plan

- [ ] Reproduce locally on a `ko-KR` iOS 26.4 simulator with the exact device (iPhone 17 Pro), cold-launch a Flutter app, capture `ax-bridge-native dump --debug` stderr verbatim
- [x] Add `--debug` or `--verbose` flag to `src/native/ax-bridge.swift` so stderr is machine-readable and the wrapper can surface it cleanly _(landed via PR #685 `cf641a9f` + follow-up `d5c0cfc0`; emits one JSON object per line on stderr, stdout payload byte-unchanged)_
- [ ] If SpringBoard is not being visited by the walk, add an explicit `bundleId=com.apple.springboard` entry-point path and union its subtree with the currently-frontmost-app walk
- [ ] Verify `app_handle_alert { action: "dismiss" }` returns `strategy: 'ax-scan'` (not `keyboard-fallback`) with visibleButtons populated
- [x] Add integration test gated by a ko-KR simulator availability check in CI

## Exit criteria

[...]

PR #658, opened 2026-04-23
fix(app-screenshot-native): retry transient surface timeout, drop informational stderr (#651)

shaun0927:
## Summary
- Retry `simctl io … screenshot` up to 2 times with 1.5s backoff when it emits `Timeout waiting for screen surfaces` — a known-transient error after high-velocity input events while the simulator composites.
- Strip the informational `Note: No display specified.` line from stderr before surfacing it as an error so operators see the actual failure signal.
- Surface a `retries` counter in the response metadata for diagnostics.

Addresses bug #2 from #651.

## Test plan
- [x] `npx jest tests/unit/app-screenshot-native.test.ts` — 10/10 pass
- [x] `npm run build` — green
- [ ] Manual: stress swipe loop (5× `app_swipe_native distance=2500`) + `app_screenshot_native` loop — expect ≥0 absorbed retries (reflected in `retries` metadata), all calls succeed

## Notes for reviewer
- No behavior change on the happy path: a first-attempt success still returns immediately with `retries: 0`.
- Non-transient simctl errors (`Invalid device state: Shutdown`, unknown device UDID, etc.) are not retried.
- Retry budget is deliberately small (2 retries × 1.5s = 3s) so a persistently broken simulator still fails fast rather than hanging the caller.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #686, opened 2026-04-27
feat(ax-bridge): bound AX messaging at 1.5s per element (#660)

shaun0927:
## Summary
- Wrap every `AXUIElement` entering the bridge with `AXUIElementSetMessagingTimeout(element, 1.5)` so a degraded AX server on the target process can't hang individual attribute reads for the framework's 6 s default.
- Coverage:
  - Application root from `AXUIElementCreateApplication(pid)` is bound immediately in `main()` before any attribute read.
  - `getChildren()` applies the same bound to every child it returns — the entire descendant walk (windows, content root, every element visited by `buildNode`, every element resolved by `findMatchingWindow` which itself uses `getChildren`) inherits the timeout transitively.
- Hangs against an unresponsive process now surface as `kAXErrorCannotComplete` (-25204) quickly instead of accumulating into the silent non-zero exits #651 described.

## Why
Implements the third bullet in m13v's recommendation on #660:
> wrap every AX call in AXUIElementSetMessagingTimeout at 1-2s

`AXUIElementSetMessagingTimeout` is per-element and does **not** propagate from a parent to children, which is why the bound has to be applied wherever an element first enters the bridge.

## PR sequence

This is the second of three independent PRs cleaving the work in #660:

| PR | Scope | Status |
|---|---|---|
| A | `--debug` flag — diagnostic only | #685 |
| B | Wrap AX calls with `AXUIElementSetMessagingTimeout` | this PR |
| C | SpringBoard explicit AXApplication union walk | later (depends on A's stderr instrumentation to characterise the failure tree) |

PR A and PR B are independent — they touch different lines and can land in any order.

## Test plan
- [x] `swiftc -O src/native/ax-bridge.swift -o dist/ax-bridge-native` — compiles cleanly
- [x] `dist/ax-bridge-native dump --device <udid> --max-depth 2` against a booted `iPhone 17 Pro` (no permission sheet visible) — returns the same JSON shape as the pre-PR baseline **byte-for-byte** on the happy path
- [x] `dist/ax-bridge-native dump --device any` with no booted simulator — exits with the existing `SIMULATOR_NOT_RUNNING` JSON, no regression
- [ ] Reviewer manual: run against a degraded simulator (background it briefly so the AX server enters the documented `kAXErrorCannotComplete` state) — confirm the dump returns within ~1.5 s instead of hanging on the previous 6 s default

## Notes for reviewer
- The timeout only manifests on degraded paths. On the happy path the AX server replies well under 1.5 s, so this PR is a no-op for typical dump / query / inspect / press calls.
- 1.5 s was chosen to be slightly under the 1–2 s window m13v recommended, leaving headroom for the wrapper's own retry/backoff before the framework's default kicks in. If the reviewer prefers 1.0 s or 2.0 s, the constant `AX_MESSAGING_TIMEOUT_SECONDS` is a single edit at the top of the helpers section.
- No new external symbols. The `setAxMessagingTimeoutSafe` helper is module-private and the timeout constant is too. The behaviour is fully contained in `src/native/ax-bridge.swift`.
- This PR does **not** add the SpringBoard explicit walk (PR C). That walker change requires live AX-tree data from a `ko-KR` cold-launch with the omofictions push-permission sheet visible to confirm where the sheet renders in the Simulator.app window hierarchy. PR A's stderr instrumentation is the prerequisite for capturing that data; PR C will land afterwards.

## Related
- Parent issue: #660
- Originating bug: #651 (closed; TS-side workarounds in #658 and #659 already shipped)
- Sibling PR: #685 (`--debug` flag)

### DIFF
--- a/src/native/ax-bridge.swift
+++ b/src/native/ax-bridge.swift
@@ -127,6 +127,34 @@ struct PressResponseJSON: Codable {
 
 // MARK: - AXUIElement Helpers
 
+// Issue #660: Per-element messaging timeout for every AXUIElement the
+// bridge touches. Without this, `AXUIElementCopyAttributeValue` against
+// an unresponsive system process (SpringBoard hosting a permission
+// sheet, a Simulator window whose AX server has degraded) blocks for
+// the AX framework's 6 s default — long enough that the wrapper's own
+// timeout fires first and the bridge looks like it has hung silently.
+//
+// Setting a 1.5 s timeout makes those blocks return
+// `kAXErrorCannotComplete` (-25204) which the call sites already treat
+// as "attribute unavailable", so the dump returns a partial tree
+// quickly instead of dragging the whole pipeline. The TS wrapper then
+// surfaces the partial result and the caller can fall back to its
+// keyboard-fallback path (#659) or its retry policy.
+//
+// `AXUIElementSetMessagingTimeout` is per-element: it does NOT
+// propagate from a parent to children, so we apply it everywhere a new
+// element enters the bridge — at the application root in `main()` and
+// at every child returned by `getChildren()`. The matched-window and
+// content-root paths are covered transitively because both arrive via
+// `getChildren()` walks.
+let AX_MESSAGING_TIMEOUT_SECONDS: Float = 1.5
+
+@discardableResult
+func setAxMessagingTimeoutSafe(_ element: AXUIElement,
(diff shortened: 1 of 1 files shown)


shaun0927:
CI status note: the failing `test` job is caused by a pre-existing bug on `develop` (`tests/unit/device-network.test.ts:379` — `getHandler` not declared, TS2552; fix queued in #679). This PR's diff does not touch the device-network suite or any TypeScript test infrastructure. Once #679 lands the test job here will pass without further changes.

PR #690, opened 2026-04-28
Rescue coordinate taps with AX button press

shaun0927:
## Summary
- Extends `app_tap`'s AX rescue path beyond modal buttons.
- If a non-raw coordinate falls inside an enabled `AXButton` frame, `app_tap` now invokes AXPress instead of relying only on OS-level coordinate injection.
- Keeps existing modal radius snapping and `raw: true` coordinate-only opt-out.
- Adds unit coverage for the non-modal AXButton rescue path.

## Why
While validating Omofictions-App App Review issues, `app_tree` could reliably see Flutter semantics buttons, but `app_tap`/SimHID coordinate taps returned success without changing the app state and often failed post-tap context probing. PointerService probing also returned success without delivering the Flutter tap. The target controls were already exposed as actionable `AXButton` nodes, so a semantic AXPress fallback is the safer headless behavior.

## Verification
- `npm test -- --runTestsByPath tests/unit/app-tap-safety.test.ts --runInBand`
- `npm run lint -- --quiet`

## Notes
I did not live-retest through the Codex MCP host after this change because that host is already running the previously built OpenSafari server. The code path is covered at the tool unit level and should take effect after build/restart.

### DIFF
--- a/src/tools/app-tap.ts
+++ b/src/tools/app-tap.ts
@@ -14,7 +14,10 @@
  *   * When a modal (AXSheet / AXDialog / AXAlert) is in the pre-tap AX
  *     tree, `app_tap` snaps the coordinate to the nearest enabled button
  *     frame and calls the AX press path instead of forwarding the raw
- *     coordinate. Callers can opt out with `raw: true`.
+ *     coordinate. For non-modal screens, coordinates that fall inside an
+ *     enabled AXButton frame use the same AX press path so headless Flutter
+ *     GestureDetector/semantics buttons are still actionable when OS-level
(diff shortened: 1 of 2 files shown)


shaun0927:
Addressed Gemini feedback in commit bf4af9f: non-modal AXButton rescue now prefers the smallest containing button frame, with a nested-button unit test covering the selection heuristic. Re-ran `npm test -- --runTestsByPath tests/unit/app-tap-safety.test.ts --runInBand` and `npm run lint -- --quiet` locally.

PR #692, opened 2026-04-29
test(integration): gated ko-KR push-permission live suite (#660)

shaun0927:
## Summary

- Issue #660 plan item: "Add integration test gated by a ko-KR simulator availability check in CI". Encodes the three Exit criteria from the issue body as Jest assertions.
- Two tests in `tests/integration/issue-660-ko-kr-push-permission.live.test.ts`:
  - **Test A** — reset → cold-launch → `app_handle_alert { action: 'dismiss' }` returns `dismissed: true, strategy: 'ax-scan'`, and `visibleButtons` contains `허용` / `허용 안 함`.
  - **Test B** — `ax-bridge-native dump --debug` exits 0 and emits the walker debug events introduced in #691.
- Gated by **explicit opt-in**: `OPENSAFARI_KOKR_PUSH_DIALOG=1` + `OSF_DEVICE_ID` + `OSF_PERMISSION_BUNDLE`. Default behavior (no opt-in) is a single \"suite skipped\" assertion that prints a structured warning. CI does not set the opt-in env var so this PR does not change CI behavior.

## Why opt-in instead of auto-detect

The cold-launch step requires `xcrun simctl privacy <udid> reset notifications <bundle>`, which fails with `NSPOSIXErrorDomain code 1` (\"Operation not permitted\") on hosts whose shell does not have Full Disk Access for the simulator's TCC sandbox — the same wall the previous autopilot session and this one both hit. Auto-running the suite on a non-FDA host would surface the simctl error as a hard test failure even though the failure is environmental, not regression-driven.

The opt-in env var doubles as a contract: the operator who flips it accepts responsibility for ensuring the host environment is FDA-granted (recipe in `docs/recipes/transient-simctl-errors.md`).

## Dependency on PR #691

Test B asserts the presence of `walker_app_windows_enumerated`, `walker_top_candidates`, and `walker_overlay_roles_seen` events. Those events ship in #691. Opt-in runs against a develop branch that includes #691 pass cleanly. Opt-in runs against an earlier develop will hit a partial-fail on Test B's event-name assertions; default-skip runs are unaffected because the assertions only execute inside the opt-in branch.

## Test plan

- [x] Default-skip run (no opt-in env var): `npx jest tests/integration/issue-660-ko-kr-push-permission.live.test.ts --runInBand` passes 1 test with the structured skip warning
- [x] ESLint clean
- [ ] Opt-in run on an FDA-granted ko-KR host: deferred until a focused session with the FDA grant runs. The suite is the verification surface, not a precondition.

### DIFF
--- a/tests/integration/issue-660-ko-kr-push-permission.live.test.ts
+++ b/tests/integration/issue-660-ko-kr-push-permission.live.test.ts
@@ -24,6 +24,12 @@
  * That keeps CI green on hosts that lack the FDA grant while preserving a
  * one-line opt-in for any focused session that does have the grant.
  *
+ * Once opt-in is set, environmental failures (TCC reset failure, missing
+ * sheet, empty tool body) throw rather than soft-skipping. The operator
+ * who flipped the opt-in env var has committed to having a working
+ * environment, so a degraded environment is a setup error that must be
+ * surfaced rather than masked.
+ *
  * Run:
  *   OPENSAFARI_KOKR_PUSH_DIALOG=1 \
  *   OSF_DEVICE_ID=<UDID-of-ko-KR-iOS-26.4-sim> \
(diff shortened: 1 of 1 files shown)


shaun0927:
### Bot review iteration summary

All four review items from the initial round on `319edbc0` are addressed in the two follow-up commits on this branch:

| Item | Source | Commit |
|---|---|---|
| `simctl` reset failure should throw, not warn (gemini@:204) | `d9f1599c` | line 244: `throw new Error(\`[issue-660] ${reset.reason}\`)` |
| Empty tool body should throw (gemini@:232) | `d9f1599c` | line 271: `throw new Error('[issue-660] tool handler returned no body')` |
| Missing permission sheet should throw (gemini@:246) | `d9f1599c` | line 284: `throw new Error(...)` |
| `visibleButtons` may carry `(norm: …)` annotation — strip before comparing (codex P1@:253) | `d9f1599c` | `stripDiagnosticAnnotation` + `canonicalLabel` mirror the regex from `src/tools/app-handle-alert.ts` |
| `\s` regex misses U+2060 WORD JOINER (codex P2@:80, new on `d9f1599c`) | `b71c6c95` | `FANCY_WHITESPACE = /[   ⁠  ]/g` mirrors `app-handle-alert-labels.ts` 1:1 |

The skip-with-warn pattern is preserved for the **default no-opt-in path** only. Once `OPENSAFARI_KOKR_PUSH_DIALOG=1` flips, every environmental failure throws.

Re-posted bot comments on `b71c6c95` cite the same line numbers as the addressed items but the actual code at those lines is the throw / canonicalize form requested — likely a bot artifact from re-evaluating the new commit. Confirmed by reading the file post-`b71c6c95`.

CI: lint + test + build all green on `b71c6c95`.

Issue #693, opened 2026-04-29
iPhone 17 Pro simulator taps use wrong coordinate space and AX post-tap probe fails


shaun0927:
## Summary

OpenSafari's `app_tap_element` and `app_double_tap` cannot reliably drive a Flutter app on an iPhone 17 Pro / iOS 26.4 simulator. Two failure surfaces are entangled and both must be closed before the App Review verification flow (settings, notification opt-in, IAP, account deletion, UGC controls) can run autonomously:

1. **AX dump empty on Flutter sub-screens** (e.g. the in-app `Ducats Shop` route). `ax-bridge-native dump` exits non-zero with `DEVICE_CONTENT_ROOT_EMPTY`; the matched device window has descendants but every candidate the walker scores reports `appSemanticsCount: 0`.
2. **Coordinate space mismatch** between the AX frames returned by the bridge (in macOS-screen-points relative to the device-content-root, ≈ 697×1515 on this combination) and the iOS-points the SimHID dispatcher consumes (402×874 on iPhone 17 Pro). Even when the AX tree is populated, taps issued at the AX-frame center can land outside the visible target.

## Environment

- App under test: Flutter app `com.omofictions.omofictionsApp`
- Simulator: iPhone 17 Pro, iOS 26.4
- Device UDID: `D7D26213-C3E9-4623-BCCB-984CDF5D0793`
- OpenSafari binary: `/Users/jh0927/opensafari/dist/ax-bridge-native`
- Date: 2026-04-29 KST

## Reproduction

Tested live in this session against the booted device above. PR #691's walker `--debug` events confirm both surfaces; quoted output is verbatim from the failing `dump` invocation.

### Surface A — AX dump empty on the Ducats Shop screen

1. Cold-launch the app: `xcrun simctl launch <UDID> com.omofictions.omofictionsApp`. App restores to the in-app `Ducats Shop` route.
2. Run `ax-bridge-native dump --device <UDID> --max-depth 12 --debug`.
3. Bridge exits 1 with `DEVICE_CONTENT_ROOT_EMPTY`. The new walker debug events from #691 show the precise failure shape:
   ```
   walker_app_windows_enumerated count=2 (iPhone 17 Pro device window + AXMenuBar)
   walker_top_candidates totalCandidates=8, all 5 top entries have appSemanticsCount=0
     [0] depth=1 role=AXGroup score=5 (matches iOSContentGroup trait)
[...]

PR #694, opened 2026-04-29
fix(ax-bridge-wrapper): surface structured ErrorJSON from STDOUT (#693 WU1)

shaun0927:
## Summary

Issue #693 WU1: the Swift bridge writes its structured ErrorJSON (`{ error, code }`) to **STDOUT** and then `exit(1)`, NOT to stderr — see `outputError()` in `src/native/ax-bridge.swift`. The wrapper's non-zero-exit branch only inspected `error.stderr`, so every typed bridge error (`DEVICE_CONTENT_ROOT_EMPTY`, `DEVICE_RESOLUTION_FAILED`, `SIMULATOR_NOT_RUNNING`, `DEVICE_WINDOW_NOT_FOUND`, etc.) collapsed to the generic `BRIDGE_EXEC_FAILED` shape with a `Command failed: <cmd>` tail and the caller could not branch on `code`.

The user-reported repro on iPhone 17 Pro / iOS 26.4 (#693) showed exactly this — every `app_tree` / `app_query` / `app_tap_element` diagnostic from the omofictions session read `Command failed: ... dump --device ... --max-depth 6` and gave no hint that the underlying condition was `DEVICE_CONTENT_ROOT_EMPTY`.

## Changes

- `src/native/accessibility-bridge.ts:exec()` non-zero-exit branch now inspects `error.stdout` first for parseable ErrorJSON; falls back to `error.stderr` for legacy paths (and `swift` interpreter compile errors that emit JSON on stderr).
- When neither stream parses as ErrorJSON, the thrown message includes both stdout and stderr tails so the caller sees the same diagnostic shape regardless of which stream the bridge wrote to.
- `AX_TIMEOUT` (killed-process) handling unchanged.

This is a strictly additive change to the error-surfacing layer. Stdout parsing on the success path is untouched. No call site assumes `BRIDGE_EXEC_FAILED` for typed-error responses, so this fix does not break callers that branch on `code` — it gives them the structured codes they were always supposed to see.

## Why this is WU1 of #693

Issue #693 has two entangled failure surfaces (AX-empty Flutter sub-screens + AX-frame to iOS-point coordinate mismatch). Localising either requires the wrapper to surface the bridge's typed error code rather than `Command failed: …`. WU1 unblocks all subsequent work units; WU2-WU5 in the issue body cover the substantive Flutter Semantics activation + coordinate conversion fixes.

## Test plan

- [x] 4 new unit tests in `tests/unit/native-accessibility-bridge.test.ts`:
  - structured ErrorJSON from STDOUT (`DEVICE_CONTENT_ROOT_EMPTY` round-trip)
  - fallback parse from STDERR (`SIMULATOR_NOT_RUNNING` legacy shape)
  - unstructured both-stream output (BRIDGE_EXEC_FAILED with both tails)
  - killed-process (`AX_TIMEOUT` preserved)
- [x] `npx jest tests/unit/` — 2187 tests pass across 149 suites
- [x] `npx eslint src/native/accessibility-bridge.ts tests/unit/native-accessibility-bridge.test.ts` — clean

### DIFF
--- a/src/native/accessibility-bridge.ts
+++ b/src/native/accessibility-bridge.ts
@@ -18,6 +18,35 @@ const DEFAULT_MAX_DEPTH = 10;
 const DEFAULT_MAX_RESULTS = 50;
 const DEFAULT_TIMEOUT_MS = 15_000;
 
+/**
+ * Maximum characters retained per line in the diagnostic tails on the
+ * `BRIDGE_EXEC_FAILED` path. The bridge runs with `maxBuffer: 10 MB` so a
+ * degenerate single-line output could otherwise produce a multi-megabyte
+ * error message. 512 chars is roughly one terminal screen and enough to
+ * carry a stack frame, a `swiftc` diagnostic, or a JSON fragment without
+ * blowing log volume.
+ */
+const STREAM_TAIL_MAX_LINES = 5;
+const STREAM_TAIL_MAX_LINE_LENGTH = 512;
+
+function formatStreamTail(label: string, value: string | undefined): string {
+  if (!value) return '';
+  const trimmed = value.trim();
+  if (trimmed.length === 0) return '';
+  const truncated = trimmed
+    .split('\n')
+    .slice(-STREAM_TAIL_MAX_LINES)
+    .map((line) =>
+      line.length > STREAM_TAIL_MAX_LINE_LENGTH
+        ? `${line.slice(0, STREAM_TAIL_MAX_LINE_LENGTH)}…[+${
+            line.length - STREAM_TAIL_MAX_LINE_LENGTH
+          } chars]`
+        : line,
+    )
+    .join(' / ');
+  return ` | ${label}: ${truncated}`;
+}
+
 export interface AccessibilityBridgeOptions {
   /**
    * Pre-resolve the bridge path, bypassing the filesystem search.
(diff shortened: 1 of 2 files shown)


shaun0927:
### Re: gemini @:180 (SyntaxError diagnostic-tail loss)

The latest commit (\`4a945f9f\`) addresses this exact concern. The wrapper now captures \`{ stdout, stderr }\` into closure variables \`capturedStdout\` / \`capturedStderr\` on the resolve path **before** \`JSON.parse\` runs (lines 156-158), and the catch block reads them via \`stdoutForDiag = error.stdout ?? capturedStdout\` (lines 191-193). This covers both rejection paths:

- **Non-zero exit**: Node populates \`error.stdout\` / \`error.stderr\` directly.
- **Successful exit + malformed stdout**: Node throws \`SyntaxError\` (which has no stdout/stderr fields), the catch block reads from the closure variables, and the BRIDGE_EXEC_FAILED message includes both stdout and stderr tails.

Verified by the unit test \"surfaces captured stdout in error message when JSON.parse throws SyntaxError on successful exit\" in \`tests/unit/native-accessibility-bridge.test.ts\`. Bot review appears to have re-evaluated the new commit and re-flagged based on the same line range without picking up the closure-variable fix above.

PR #695, opened 2026-04-29
feat(ax-bridge): emit deviceContentMacOSPt on dump root (#693 WU3-prep)

shaun0927:
## Summary

Issue #693 Surface B is the AX-frame to iOS-point coordinate-space mismatch (observed at 1.733× on iPhone 17 Pro / iOS 26.4). The bridge already returns frames in macOS-screen-points relative to the device-content-root origin, and `sim-hid-bridge` already consumes iOS-points — what was missing was the size of the device-content-root in macOS-points so the TS-side conversion factor can be derived without a separate `simctl` runtime call per tap.

This PR ships the **data plane only**: `ax-bridge` now emits `deviceContentMacOSPt: { width, height }` on the dump root. The actual TS-side conversion in `app-tap-element.ts` lives in a follow-up WU3-proper PR.

## Changes

- `AXNodeJSON` gains an optional `deviceContentMacOSPt: SizeJSON?` field, populated **only on the dump root** (mirrors the existing `chromeOnly` pattern — per-child nodes do not carry the noise).
- `main()` reads `getSize(content)` once after `findDeviceContentRecursively` succeeds and assigns the result onto the dump root before `outputJSON`. New `--debug` event `device_content_macos_pt` surfaces the captured size; `device_content_macos_pt_unavailable` fires when `getSize` returns nil (extremely rare; `kAXErrorCannotComplete` against a degraded AX server).
- TS `AXNode` type gets the optional `deviceContentMacOSPt` field for type-safe consumption.
- Two new unit tests in `tests/unit/native-accessibility-bridge.test.ts`: passthrough on dump root, legacy-bridge backward compatibility (older binaries / interpreter source produce no field — wrapper must keep working).

## Why ship this independently of WU3-proper

WU3-proper still needs the iOS-points device size, and the design choice between (a) a dedicated `simctl` runtime lookup, (b) propagating `sim-hid-bridge`'s `getScreenSize` through TS, or (c) a new `--input-space=macos-pt` flag on `sim-hid-bridge` that does the conversion internally is open. Landing the data plane now means any of those choices can be implemented in a follow-up without another `ax-bridge` invocation per tap.

## Test plan

- [x] `swiftc -O src/native/ax-bridge.swift` compiles cleanly
- [x] `npx jest tests/unit/` — 2185 tests pass across 149 suites
- [x] `npx eslint` — clean
- [ ] Live verification with a populated AX tree: blocked in this session by the same Simulator AX-server degradation #693 reports (the AX tree is empty across multiple foreground apps including Mobile Safari / Settings / the omofictions Flutter app). The change is purely additive and bypasses the AX-empty path entirely (`getSize` is called only after `findDeviceContentRecursively` returns non-nil).

### DIFF
--- a/src/native/ax-bridge.swift
+++ b/src/native/ax-bridge.swift
@@ -88,6 +88,13 @@ struct QueryResultJSON: Codable {
     // results into typed APP_CONTENT_NOT_EXPOSED instead of issuing a
     // second native dump call.
     let chromeOnly: Bool
+    // Issue #693 WU3-prep (gemini PR #695 follow-up): emit the device-
+    // content-root size in macOS-pt on query results too, so a caller that
+    // uses `query` to find an element and then performs a coordinate-based
+    // tap does not need to issue a separate `dump` just to retrieve the
+    // conversion factor. Optional for forward compatibility with older
+    // wrappers.
+    let deviceContentMacOSPt: SizeJSON?
 }
 
 struct QueryJSON: Codable {
@@ -107,12 +114,18 @@ struct ErrorJSON: Codable {
 /// `chromeOnly` flag the wrapper relies on for query/dump promotion. The
 /// wrapper inspects this field to decide whether to upgrade the response
 /// to APP_CONTENT_NOT_EXPOSED.
+///
+/// Issue #693 WU3-prep (gemini PR #695 follow-up): also carries
+/// `deviceContentMacOSPt` so a caller that wants to retry the operation
+/// at coordinate level (after a chrome-only or transient miss) has the
+/// conversion factor without issuing a separate dump.
 struct InspectNotFoundJSON: Codable {
     let error: String
     let code: String
     let path: String
     let found: Bool
     let chromeOnly: Bool
+    let deviceContentMacOSPt: SizeJSON?
 }
 
 /// Uniform response for `ax-bridge press`.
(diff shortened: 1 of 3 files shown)


shaun0927:
### Re: gemini @:35 and :1120 (propagate to query/inspect)

Latest commit \`5c8c8d22\` addresses both. The current Swift bridge:

- \`QueryResultJSON\` declares \`deviceContentMacOSPt: SizeJSON?\` (line 91 in the diff)
- \`InspectNotFoundJSON\` declares \`deviceContentMacOSPt: SizeJSON?\` (line 128 in the diff)
- \`main()\` query case constructs the result with \`deviceContentMacOSPt: deviceContentSizeJSON\`
- \`main()\` inspect not-found path constructs \`InspectNotFoundJSON\` with \`deviceContentMacOSPt: deviceContentSizeJSON\`
- \`main()\` inspect success path assigns \`node.deviceContentMacOSPt = deviceContentSizeJSON\` before \`outputJSON\`

TS \`AXQueryResult\` interface also gained the optional \`deviceContentMacOSPt\` field, and two new unit tests verify query/inspect passthrough.

The bot review appears to have re-evaluated the new commit and re-flagged based on the same line range without picking up the propagation already in place. CI is green.

Issue #705, opened 2026-05-06
Reduce process-spawn overhead in native input routing


shaun0927:
Tracked by #713.

## Background
Headless native input is a major OpenSafari differentiator. Repeated shelling out for taps, swipes, key input, AppleScript, or Swift snippets can slow input-heavy workflows.

## Objective
Improve native input throughput by preferring persistent or batched execution paths where stable backends exist.

## Scope
- Identify high-frequency process-spawn input paths.
- Prefer compiled bridge binaries over interpreter snippets when available.
- Add a batching API for repeated input operations.
- Keep AppleScript fallback opt-in as the documented last resort.

## Target files and paths
- `src/tools/native-input-backend.ts`
- `src/tools/sim-hid-input-backend.ts`
- `src/tools/pointer-service-input-backend.ts`
- `src/native/*.swift`
- `tests/unit/native-input-backend.test.ts`
- `docs/headless-architecture.md`

## Implementation steps
1. Measure current process-spawn points for tap, swipe, type, and key operations.
2. Route stable backend calls through compiled bridges when present.
3. Design a minimal batch operation shape for repeated taps/keys.
4. Cache recent Simulator activation state for focus-stealing fallback only.
5. Add unit tests for backend selection and batching behavior.

## Acceptance criteria
- [x] Input-heavy workflow can execute repeated operations without one process spawn per logical event where a stable backend supports batching.
- [x] AppleScript remains opt-in and is not promoted to default.
- [x] Backend fallback order remains documented and tested.
- [x] Compiled bridge fallback behavior is preserved when binaries are unavailable.

## Verification
- `npm run lint -- --quiet`
- `npm test -- --runInBand tests/unit/native-input-backend.test.ts`
- Manual macOS smoke for one Safari tap and one native app element tap.

## Dependencies / recommended order
Run after readiness and socket startup issues. This issue may be split by backend if the first implementation exceeds one PR.

## Risks and cautions
- Do not change experimental backend stability labels without updating docs and tests.
- Do not make focus-stealing input default.
- Do not remove working fallback paths.

## Non-goals
- Do not solve Xcode 26 coordinate tap research in this issue.
- Do not redesign all app interaction tools.
- Do not require persistent daemon installation.

## Ambiguity reducers
- “Batching” means grouping repeated input actions in one backend invocation when semantics allow it.
- A backend must advertise batching support; callers must not assume every backend supports it.
- AppleScript activation caching applies only to the opt-in fallback path.

## Validity review
[...]

PR #724, opened 2026-05-07
perf(simulator): bootstatus-aware polling with shared state cache (#703)

### LINKED ISSUE #703: Reduce simulator lifecycle polling overhead in multi-device workflows
Tracked by #713.

## Background
Parallel simulator automation is a core OpenSafari value. Repeated full `simctl list devices -j` polling during boot/shutdown can become expensive on machines with many runtimes or devices.

## Objective
Make simulator lifecycle waits cheaper and more deterministic for parallel device workflows.

## Scope
- Use narrower simulator status commands where available.
- Add short-lived device state caching during polling loops.
- Share state lookups between manager and pool code.
- Keep behavior compatible with supported Xcode versions.

## Target files and paths
- `src/simulator/manager.ts`
- `src/simulator/sim-pool.ts`
- `src/simulator/pool.ts`
- `src/simulator/simctl.ts`
- `tests/unit/*simulator*.test.ts`

## Implementation steps
1. Audit current boot/shutdown polling callers.
2. Introduce a simulator state reader with a short TTL for repeated polling.
3. Use `simctl bootstatus <udid> -b` for boot readiness when it is available.
4. Keep fallback to current polling for unsupported environments.
5. Add tests for cache hit, cache invalidation, and fallback paths.

## Acceptance criteria
- [x] Boot wait does not repeatedly parse the full device list when a narrower command is available.
- [x] Shutdown wait uses a bounded polling strategy with clear timeout behavior.
- [x] Parallel pool operations share state reads during a polling tick.
- [x] Fallback behavior works when `bootstatus` is unavailable.

## Verification
- `npm run lint -- --quiet`
- Targeted simulator manager/pool unit tests
- Manual macOS smoke: boot and shutdown one simulator, then two simulators in parallel.

## Dependencies / recommended order
Run after the proxy/WebKit speed issues or in parallel by a separate PR because it touches simulator lifecycle only.

## Risks and cautions
- Do not assume all Xcode versions support the same `simctl` commands.
- Do not cache state across lifecycle mutations without invalidation.
- Do not hide timeout errors.

## Non-goals
- Do not redesign device preset selection.
- Do not change public tool names or parameters.
- Do not optimize native input routing in this issue.

## Ambiguity reducers
- “Short TTL” means cache lifespan measured in a single polling interval, not long-term device inventory caching.
- Fallback must be explicit and tested.
- Lifecycle actions invalidate relevant cache entries.

## Validity review
Proceed. This directly supports parallel multi-device automation and CI speed.

## Implementation detail refinements
- Wrap `bootstatus` use in a capability-detected helper.
- Record the command path in debug logs without excessive noise.
- Prefer dependency injection for `Simctl` in tests.

## Verification checklist reinforcement
- [x] Boot path uses `bootstatus` or documented fallback.
- [x] Cache invalidates after boot/shutdown/delete.
- [x] Unit tests cover unavailable-command fallback.
- [x] Parallel boot test avoids duplicate full-list calls in mock assertions.

shaun0927:
## Summary

Closes #703

Reduces `simctl` overhead in multi-device polling loops by introducing a short-lived `SimulatorStateCache` (TTL ≈ one polling interval) so concurrent callers within the same poll tick share a single state read.

> **Note:** an earlier iteration of this PR also routed polling through `simctl bootstatus <udid> -b` for a narrower command. That optimization was **dropped** during review — `bootstatus -b` is a mutating command (it boots the device when called against a Shutdown simulator) and therefore unsafe for use as a state probe (Codex P2 finding). The dedup-via-cache change alone delivers the goal of #703 without compromising the read-only contract of `getDeviceState()`.

## Boot/shutdown polling callers (audit)

Before this PR, all three callers issued a full `simctl list devices` parse on every tick:

| Caller | File | Poll interval |
|---|---|---|
| `SimulatorManager.boot()` | `src/simulator/manager.ts` | 1 000 ms |
| `SimulatorManager.shutdown()` | `src/simulator/manager.ts` | 1 000 ms |
| `SimPool.waitForBootedState()` | `src/simulator/sim-pool.ts` | 500 ms |

All three now share the cache via `manager.getDeviceState()`. The first caller in a tick performs the lookup (per-UDID `simctl list devices <udid> -j`, a pure read); subsequent callers within the TTL receive the cached snapshot.

## Implementation notes

- **`SimulatorStateCache`** — TTL matched to the longest poll interval (900 ms) so the next tick always fetches fresh data; not a long-term inventory cache.
- **Per-UDID list read** — `getDeviceState()` uses `simctl list devices <udid> -j`, which scopes the parse to a single device record and is side-effect-free.
- **Cache invalidation** — `boot()`, `shutdown()`, and `deleteDevice()` all call `stateCache.invalidate(udid)` immediately after issuing the lifecycle command; no stale state can persist across mutations.
- **Timeout surfaces** — `shutdown()` timeout still reaches the nuclear erase path and does not swallow errors silently (verified by test).
- **Race-disappear handling** — if a device disappears between the state probe and metadata fetch in `boot()`, we now throw `DeviceNotFoundError` rather than silently returning the stale pre-boot snapshot.
- **Debug logging** — command path logged under `process.env.DEBUG` only, no noise in normal operation.

## Non-goals (per issue)

- No changes to device preset selection
- No changes to public tool names or parameters
- Native input routing optimisation is tracked separately in #705

## Test plan

- [x] `SimulatorStateCache` TTL, invalidation, `invalidateAll()`
- [x] Per-UDID list read returns booted/shutdown state correctly
- [x] `getDeviceState()` never invokes `bootstatus -b` (read-only contract regression test)
- [x] Cache invalidated after `boot()`, `shutdown()`, `deleteDevice()`
- [x] `shutdown()` timeout surfaces nuclear erase (not silently swallowed)
- [x] `boot()` race-disappear throws `DeviceNotFoundError` instead of returning stale Shutdown snapshot
- [x] Parallel pool acquires call `getDeviceState` (not `getDevice`) for polling tick
- [x] All existing `sim-pool.test.ts` tests still pass
- [x] All existing `warm-sim-pool.test.ts` tests still pass
- [x] `npx tsc --noEmit --pretty false` clean
- [x] `npx eslint --quiet` on touched files clean
- [x] `git diff --check` clean

### DIFF
--- a/src/simulator/manager.ts
+++ b/src/simulator/manager.ts
@@ -4,7 +4,7 @@ import * as path from 'path';
 import { randomUUID } from 'crypto';
 import { execFile } from 'child_process';
 import { promisify } from 'util';
-import { SimctlExecutor, SimctlError } from './simctl';
+import { SimctlExecutor, SimctlError, SimulatorStateCache, hasBootstatus } from './simctl';
 import { SimulatorDevice, SimulatorRuntime } from './types';
 import { DEVICE_PRESETS } from './presets';
 import { DEFAULT_SIMULATOR_BOOT_TIMEOUT_MS, DEFAULT_SIMULATOR_SHUTDOWN_TIMEOUT_MS, DEFAULT_SCREENSHOT_TIMEOUT_MS } from '../config/defaults';
@@ -30,8 +30,21 @@ export interface RotationResult {
   orientation?: string;
 }
 
+/**
+ * TTL for the per-device state cache inside SimulatorManager.
+ * Sized to match the longest polling interval used in this file (1 000 ms) so
+ * that multiple callers within a single tick share one `simctl list devices`
+ * parse, but the next tick always sees fresh data.
+ */
+const STATE_CACHE_TTL_MS = 900;
+
 export class SimulatorManager {
   private simctl = new SimctlExecutor();
+  /**
+   * Short-lived cache for device states during polling loops.
+   * Invalidated immediately on any lifecycle mutation (boot / shutdown / delete).
+   */
+  private stateCache = new SimulatorStateCache(STATE_CACHE_TTL_MS);
 
   async listDevices(): Promise<SimulatorDevice[]> {
     const result = await this.simctl.execJson<SimctlListResult>(['list', 'devices']);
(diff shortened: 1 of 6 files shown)


shaun0927:
@codex review

Re-requesting Codex review after the round-3 fix in commit `21d1667f` which addresses all three previously flagged issues:

- (Codex P2 critical premise) Dropped the mutating `bootstatus -b` fast-path entirely. `getDeviceState()` is now read-only via `simctl list devices <udid> -j`.
- (Codex P1) Removed `hasBootstatus` capability probe — dead code after the bootstatus removal.
- (Codex P2) `boot()` now throws `DeviceNotFoundError` if the post-Booted re-fetch misses, rather than returning a stale pre-boot snapshot.

shaun0927:
@codex review

PR #730, opened 2026-05-07
perf(webkit): batch navigation state read + dedupe domain enables (#702 b/2)

### LINKED ISSUE #702: Optimize WebKit navigation and evaluation hot paths
Tracked by #713.

## Background
OpenSafari’s core work is WebKit Remote Debugging Protocol automation. Navigation, screenshot, DOM inspection, and evaluate are frequent paths, so reducing protocol round trips improves real user speed.

## Objective
Add fast value-return evaluation and reduce redundant WebKit RPCs in navigation and screenshot flows.

## Scope
- Add a value-oriented evaluate helper for non-promise primitive/object results.
- Batch post-navigation state reads.
- Cache enabled protocol domains per target.
- Reduce fixed sleeps where event or state checks are sufficient.

## Target files and paths
- `src/webkit/client.ts`
- new `src/webkit/evaluate.ts` or extracted command module if refactoring begins
- `tests/unit/webkit*.test.ts` or new tests for evaluation/navigation helpers

## Implementation steps
1. Introduce `evaluateValue<T>` or an equivalent option using `returnByValue: true`.
2. Update screenshot viewport query to use the fast path.
3. Combine final URL, readyState, and status reads into one evaluation.
4. Track enabled domains to avoid duplicate `Page.enable` and `Network.enable` calls.
5. Add tests using mocked send calls to assert lower RPC count.

## Acceptance criteria
- [x] Object-returning simple evaluations can complete in one protocol call.
- [x] Navigation final state is read with one evaluation call.
- [x] Repeated navigation does not re-enable already enabled domains for the same target.
- [x] Existing public `evaluate` behavior remains compatible.

## Verification
- `npm run lint -- --quiet`
- Targeted WebKit client unit tests
- Manual smoke on macOS simulator: navigate, screenshot, read_page still work.

## Dependencies / recommended order
Can run after proxy readiness work or independently if tests use mocks. Prefer before large WebKitClient refactor so performance behavior is captured by tests.

## Risks and cautions
- Do not break promise-returning evaluation semantics.
- Do not use Chrome CDP APIs; preserve WebKit protocol constraints from `CLAUDE.md`.
- Do not remove existing error handling paths.

## Non-goals
- Do not rewrite all `WebKitClient` responsibilities in this issue.
- Do not introduce Playwright/CDP abstractions.
- Do not change screenshot protocol from `Page.snapshotRect`.

## Ambiguity reducers
- “Fast path” applies only when callers expect serializable values.
- The existing generic `evaluate` remains available for promise/objectId cases.
- RPC count assertions should focus on specific helper calls, not entire workflows with simulator variability.

## Validity review
Proceed. It is central to the automation product because most user-visible actions pass through WebKit RPCs.

## Implementation detail refinements
- Add helper first, then migrate one or two hot paths per PR.
- Prefer typed return helpers over adding more `any` casts.
- Use WebKit event support only after confirming protocol availability in tests or live smoke.

## Verification checklist reinforcement
- [x] Screenshot viewport query uses value fast path.
- [x] Navigation final state uses one combined evaluation.
- [x] Tests prove public API output shape is unchanged.
- [x] No CDP-only APIs are introduced.

shaun0927:
**Stacked on \`perf/702a-evaluate-value\`** — base=\`perf/702a-evaluate-value\`. Will retarget to \`develop\` after #702a merges.

Implements #702 part (b). Closes #702 when combined with #702a.

### \`src/webkit/client.ts\`
- **Batch navigation final state read** (\`navigate()\`): replaced 3 separate \`evaluate()\` calls for \`finalReadyState\`, \`currentUrl\`, and \`status\` with a single \`evaluateValue<{url, readyState, status}>(...)\` call. RPC delta: was up to 6 RPCs (3× Runtime.evaluate + up to 3× callFunctionOn for object serialisation); now exactly **1 Runtime.evaluate** with \`returnByValue:true\`.
- **\`enableDomainForTarget()\` dedup**: added a per-target Set check before issuing \`\${domain}.enable\`. Second call for the same domain+target is a no-op (no protocol round-trip). \`enableDomain()\` (global) already had this guard; \`enableDomainForTarget()\` now has the same guarantee.
- Target destruction (\`Target.targetDestroyed\`) already clears \`enabledDomainsPerTarget\` for that target, ensuring a new target for the same id re-enables cleanly.

### \`tests/unit/webkit-evaluate.test.ts\`
- **Nav batch read** — 2 new tests asserting exactly 1 \`Runtime.evaluate\` (returnByValue:true) for final nav state, 0 \`callFunctionOn\` calls, correct values propagate to \`NavigateResult\`.
- **Domain enable dedup** — 5 new tests: \`enableDomain\` dedup, cross-domain independence, \`enableDomainForTarget\` dedup, per-target independence, cache clear on \`targetDestroyed\`.

## RPC count delta on navigation path

| Step | Before | After |
|------|--------|-------|
| Final readyState check | 2 RPCs | rolled in |
| Current URL read | 2 RPCs | rolled in |
| Status read | 2 RPCs | rolled in |
| **Batch final state** | — | **1 RPC** |
| **Net saving** | **6 RPCs** | **→ 1 RPC** |

## Validation

- \`npx tsc --noEmit\` — 0 errors
- Jest: **22 passed, 0 failed** (webkit-evaluate + webkit-client-list-targets suites)
- \`npm run lint -- --quiet\` — clean
- \`git diff --check\` — clean

### DIFF
--- a/src/webkit/client.ts
+++ b/src/webkit/client.ts
@@ -395,15 +395,26 @@ export class WebKitClient extends EventEmitter implements BrowserBackend {
    * No-ops if the domain is already enabled for this target (avoids duplicate RPCs).
    */
   async enableDomainForTarget(domain: string, targetId: string): Promise<void> {
-    if (!this.enabledDomainsPerTarget.has(targetId)) {
+    const hadEntry = this.enabledDomainsPerTarget.has(targetId);
+    if (!hadEntry) {
       this.enabledDomainsPerTarget.set(targetId, new Set());
     }
     const targetDomains = this.enabledDomainsPerTarget.get(targetId)!;
     if (targetDomains.has(domain)) {
       return;
     }
-    await this.sendToTarget(`${domain}.enable`, undefined, targetId);
-    targetDomains.add(domain);
+    try {
+      await this.sendToTarget(`${domain}.enable`, undefined, targetId);
+      targetDomains.add(domain);
+    } catch (err) {
+      // If the RPC fails for an invalid/closed target, we won't receive a
+      // Target.targetDestroyed cleanup — drop the freshly-created empty Set
+      // ourselves so the map does not grow unboundedly.
+      if (!hadEntry && targetDomains.size === 0) {
+        this.enabledDomainsPerTarget.delete(targetId);
+      }
+      throw err;
+    }
   }
 
   // ========== Multi-Tab Target Management ==========
--- a/src/webkit/client.ts
+++ b/src/webkit/client.ts
@@ -392,13 +392,18 @@ export class WebKitClient extends EventEmitter implements BrowserBackend {
   /**
    * Enable a domain on a specific target (tab).
    * Tracks the domain per-target so it can be re-enabled after target recreation.
+   * No-ops if the domain is already enabled for this target (avoids duplicate RPCs).
    */
   async enableDomainForTarget(domain: string, targetId: string): Promise<void> {
-    await this.sendToTarget(`${domain}.enable`, undefined, targetId);
     if (!this.enabledDomainsPerTarget.has(targetId)) {
       this.enabledDomainsPerTarget.set(targetId, new Set());
     }
-    this.enabledDomainsPerTarget.get(targetId)!.add(domain);
+    const targetDomains = this.enabledDomainsPerTarget.get(targetId)!;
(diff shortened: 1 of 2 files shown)


shaun0927:
@codex review

PR #764, opened 2026-05-17
fix(network-intercept): evict session cache on HTTP delete

### LINKED ISSUE #763: Evict NetworkInterceptor map entries on MCP session deletion (HTTP transport)
## Summary

`src/tools/network-intercept.ts` creates one `NetworkInterceptor` per `(sessionId, deviceId)` pair (after PR #762) and keeps it in a module-scoped `interceptorsByKey` map indefinitely. The map only ever inserts; it has no `.delete(key)` path tied to MCP session lifecycle. For long-running HTTP deployments (which churn `Mcp-Session-Id` headers), each new session creates a fresh interceptor that is never collected — a slow per-session memory leak.

This is the third item from Codex's review on PR #762 (https://github.com/shaun0927/opensafari/pull/762#discussion). The other two (action validation, device scoping) were addressed in #762 because the fixes are small and self-contained. This one is split out because the cleanest fix touches MCP server plumbing (subscribing to `HTTPTransport.onSessionDelete`) and is out of scope for a release catch-up PR.

## Reproduction

```bash
# Long-running HTTP server
node dist/cli/index.js serve --http 3100 --http-token <token>

# Drive a synthetic load that opens fresh MCP sessions and runs
# network_intercept at least once per session, then closes them
# (DELETE /mcp with Mcp-Session-Id). Watch RSS / heap snapshots —
# interceptorsByKey grows linearly with session count and never shrinks.
```

## Code references

- `src/tools/network-intercept.ts:9` — `interceptorsByKey = new Map<string, NetworkInterceptor>()` declared at module scope.
- `src/tools/network-intercept.ts:24-32` — `getNetworkInterceptorForSession` only inserts; the only existing cleanup is `resetNetworkInterceptorsForTest` (test-only) which clears the whole map.
- `src/transports/http.ts:65-67` — `HTTPTransport.onSessionDelete(handler)` exists but has no callers in `src/`. This is the right hook to wire from.
- `src/transports/http.ts:494-503` — session deletion calls `sessionDeleteHandler(sessionId)`.

## Proposed fix outline

1. Add `removeNetworkInterceptorForSession(sessionId: string): void` to `src/tools/network-intercept.ts` that:
   - Iterates `interceptorsByKey` keys with the prefix `${sessionId}|` (composite key shape from #762).
   - Calls `interceptor.disable(client)` if a `client` is still resolvable for the matching `deviceId`, otherwise just deletes the map entry (no client → nothing to revert).
   - Deletes every matching key.
2. In `src/mcp-server.ts`, register the cleanup on the HTTP transport's `onSessionDelete` hook when the transport is HTTP. STDIO sessions are process-scoped so no leak applies there.
3. Add unit coverage in `tests/unit/network-intercept-tool.test.ts`:
   - Multiple `(sessionId, deviceId)` pairs created, then `removeNetworkInterceptorForSession(sessionId)` called — only matching entries removed, sibling sessions untouched.
   - Assert that subsequent `getNetworkInterceptorForSession(sessionId, deviceId)` returns a fresh interceptor (proves the prior one was actually evicted, not just disabled in place).
4. Optional: emit a debug log when an interceptor is evicted with active rules, so operators notice unintended session churn.

## Out of scope for this issue

- Reworking the keying scheme — the composite `(sessionId, deviceId)` from #762 is the right shape.
- Adding STDIO-side cleanup — STDIO sessions live for the process lifetime; there is no equivalent leak vector.
- Generalising a session-scoped registry abstraction — only `network-intercept` needs this today; other tools that gain per-session caches can be migrated incrementally.

## Acceptance criteria

- [ ] `removeNetworkInterceptorForSession(sessionId)` exported from `network-intercept.ts` and unit-tested.
- [ ] Wired into the HTTP transport's `onSessionDelete` callback in `mcp-server.ts`.
- [ ] Unit test covers eviction + sibling-session isolation.
- [ ] No regression in the existing 2656-test unit suite (currently green on `develop`).

shaun0927:
## Summary
- Exposes an optional `MCPTransport.onSessionDelete` lifecycle hook through the transport abstraction.
- Wires HTTP session deletion to `removeNetworkInterceptorForSession(sessionId)` so `network_intercept` per-session rule state is evicted when streamable HTTP sessions are deleted.
- Adds unit coverage for direct cache eviction and MCPServer lifecycle wiring.

## Issue analysis / scope
Open issue #763 is still valid: HTTP transport already has a session delete callback, but `MCPServer` did not register cleanup, so the module-level network interceptor map retained per-session `NetworkInterceptor` instances after `DELETE /mcp`.

This PR intentionally keeps the scope narrow:
- It removes stale in-memory interceptor state for the deleted session.
- It preserves stdio behavior by making the transport hook optional.
- It avoids a `mcp-server -> network-intercept` import cycle by moving the cache into a side-effect-light module that both layers can share synchronously.

## Validation
- `npx jest --runTestsByPath tests/unit/network-intercept-tool.test.ts tests/unit/mcp-server-session-cleanup.test.ts --runInBand`
- `npx eslint --max-warnings=0 src/tools/network-intercept.ts src/tools/network-intercept-cache.ts src/transports/index.ts src/mcp-server.ts tests/unit/network-intercept-tool.test.ts tests/unit/mcp-server-session-cleanup.test.ts`
- `npx tsc --noEmit --pretty false`

Closes #763

### DIFF
--- a/src/mcp-server.ts
+++ b/src/mcp-server.ts
@@ -17,6 +17,7 @@ import {
   MCPMessageContext,
 } from './types/mcp';
 import { createTransport, MCPTransport, TransportMode } from './transports';
+import { removeNetworkInterceptorForSession } from './tools/network-intercept-cache';
 import { TOOL_TIERS, getToolTier } from './config/tool-tiers';
 import { BrowserBackend } from './types/browser-backend';
 import { logAuditEntry } from './security/audit-logger';
@@ -194,6 +195,10 @@ export class MCPServer {
       authToken: options.authToken,
       insecure: options.httpInsecure,
       allowedOrigins: options.allowedOrigins,
+    });
+
+    this.transport.onSessionDelete?.((sessionId) => {
+      removeNetworkInterceptorForSession(sessionId);
     });
 
     this.transport.onMessage((msg, context) => this.handleMessage(msg, context));
--- a/src/tools/network-intercept-cache.ts
+++ b/src/tools/network-intercept-cache.ts
@@ -1,0 +1,38 @@
+import { NetworkInterceptor } from '../network-interceptor';
+
+const DEFAULT_INTERCEPTOR_SCOPE = '__default__';
+const interceptorsBySession = new Map<string, NetworkInterceptor>();
+
+function resolveInterceptorScope(sessionId?: string): string {
+  return sessionId || DEFAULT_INTERCEPTOR_SCOPE;
+}
+
+export function getNetworkInterceptorForSession(sessionId?: string): NetworkInterceptor {
+  const key = resolveInterceptorScope(sessionId);
+  let interceptor = interceptorsBySession.get(key);
+  if (!interceptor) {
+    interceptor = new NetworkInterceptor();
+    interceptorsBySession.set(key, interceptor);
+  }
+  return interceptor;
+}
+
+export function removeNetworkInterceptorForSession(sessionId: string): number {
+  if (!sessionId) return 0;
+
+  let removed = 0;
+  for (const key of Array.from(interceptorsBySession.keys())) {
+    if (key === sessionId || key.startsWith(`${sessionId}|`)) {
+      interceptorsBySession.delete(key);
+      removed += 1;
+    }
+  }
+  return removed;
+}
+
+export function resetNetworkInterceptorsForTest(): void {
+  interceptorsBySession.clear();
+}
+
+/** Legacy singleton for callers that are not yet session-aware. */
+export const networkInterceptor = getNetworkInterceptorForSession(DEFAULT_INTERCEPTOR_SCOPE);
--- a/src/tools/network-intercept.ts
+++ b/src/tools/network-intercept.ts
@@ -1,29 +1,21 @@
 import { MCPServer, getWebKitClient } from '../mcp-server';
 import {
-  NetworkInterceptor,
   type InterceptorClient,
   type InterceptRule,
 } from '../network-interceptor';
+import {
+  getNetworkInterceptorForSession,
+  networkInterceptor,
+  removeNetworkInterceptorForSession,
+  resetNetworkInterceptorsForTest,
+} from './network-intercept-cache';
 
-const DEFAULT_INTERCEPTOR_SCOPE = '__default__';
-const interceptorsBySession = new Map<string, NetworkInterceptor>();
-
-export function getNetworkInterceptorForSession(sessionId?: string): NetworkInterceptor {
-  const key = sessionId || DEFAULT_INTERCEPTOR_SCOPE;
-  let interceptor = interceptorsBySession.get(key);
-  if (!interceptor) {
-    interceptor = new NetworkInterceptor();
-    interceptorsBySession.set(key, interceptor);
-  }
-  return interceptor;
-}
-
-export function resetNetworkInterceptorsForTest(): void {
-  interceptorsBySession.clear();
-}
-
-/** Legacy singleton for callers that are not yet session-aware. */
-export const networkInterceptor = getNetworkInterceptorForSession(DEFAULT_INTERCEPTOR_SCOPE);
+export {
+  getNetworkInterceptorForSession,
+  networkInterceptor,
+  removeNetworkInterceptorForSession,
+  resetNetworkInterceptorsForTest,
+};
 
 function resolveClient(deviceId: unknown): InterceptorClient | null {
   return getWebKitClient(typeof deviceId === 'string' ? deviceId : undefined);
--- a/src/transports/index.ts
+++ b/src/transports/index.ts
@@ -26,6 +26,12 @@ export interface MCPTransport {
    */
   send(response: MCPResponse): void;
 
+  /**
+   * Register cleanup that runs when a stateful transport deletes an MCP session.
+   * Stateless transports can omit this hook.
+   */
+  onSessionDelete?(handler: (sessionId: string) => void): void;
+
   /** Start listening for messages (bind port or attach readline). */
   start(): void | Promise<void>;
 
--- a/tests/unit/mcp-server-session-cleanup.test.ts
+++ b/tests/unit/mcp-server-session-cleanup.test.ts
@@ -1,0 +1,36 @@
+describe('MCPServer HTTP session cleanup', () => {
+  beforeEach(() => {
+    jest.resetModules();
+  });
+
+  it('registers network interceptor eviction for transport session deletion', async () => {
+    const onSessionDelete = jest.fn();
+    const transport = {
+      onMessage: jest.fn(),
+      onSessionDelete,
+      send: jest.fn(),
+      start: jest.fn(),
+      close: jest.fn(),
+    };
+    const removeNetworkInterceptorForSession = jest.fn();
+
+    jest.doMock('../../src/transports', () => ({
+      createTransport: jest.fn(async () => transport),
+    }));
+    jest.doMock('../../src/tools/network-intercept-cache', () => ({
+      removeNetworkInterceptorForSession,
+    }));
+
+    const { MCPServer } = await import('../../src/mcp-server');
+    const server = new MCPServer();
+
+    await server.start({ transport: 'http' });
+
+    expect(onSessionDelete).toHaveBeenCalledWith(expect.any(Function));
+    const handler = onSessionDelete.mock.calls[0][0] as (sessionId: string) => void;
+    handler('session-a');
+    await new Promise((resolve) => setImmediate(resolve));
+
+    expect(removeNetworkInterceptorForSession).toHaveBeenCalledWith('session-a');
+  });
+});
--- a/tests/unit/network-intercept-tool.test.ts
+++ b/tests/unit/network-intercept-tool.test.ts
@@ -1,4 +1,8 @@
-import { getNetworkInterceptorForSession, resetNetworkInterceptorsForTest } from '../../src/tools/network-intercept';
+import {
+  getNetworkInterceptorForSession,
+  removeNetworkInterceptorForSession,
+  resetNetworkInterceptorsForTest,
+} from '../../src/tools/network-intercept';
 
 describe('network_intercept session scoping', () => {
   beforeEach(() => resetNetworkInterceptorsForTest());
@@ -24,6 +28,29 @@ describe('network_intercept session scoping', () => {
 
     expect(a2.listRules()).toHaveLength(1);
     expect(a2.findMatchingRule('https://example.com/login')?.action).toBe('mock');
+  });
+
+  it('evicts only the selected MCP session cache on session deletion', () => {
+    const a1 = getNetworkInterceptorForSession('session-a');
+    const b = getNetworkInterceptorForSession('session-b');
+
+    a1.addRule({ urlPattern: '/a', action: 'block' });
+    b.addRule({ urlPattern: '/b', action: 'block' });
+
+    expect(removeNetworkInterceptorForSession('session-a')).toBe(1);
+
+    const a2 = getNetworkInterceptorForSession('session-a');
+    expect(a2).not.toBe(a1);
+    expect(a2.listRules()).toHaveLength(0);
+    expect(b.listRules()).toHaveLength(1);
+  });
+
+  it('reports no eviction for unknown or empty session ids', () => {
+    getNetworkInterceptorForSession('session-a').addRule({ urlPattern: '/a', action: 'block' });
+
+    expect(removeNetworkInterceptorForSession('missing')).toBe(0);
+    expect(removeNetworkInterceptorForSession('')).toBe(0);
+    expect(getNetworkInterceptorForSession('session-a').listRules()).toHaveLength(1);
   });
 
   it('clear/disable restores and clears only the selected session', async () => {


shaun0927:
# PR Review Summary

## Verdict
Approve

## Scope Reviewed
- PR intent: close #763 by evicting per-session network interceptor cache entries when HTTP MCP sessions are deleted.
- Main changed areas: transport lifecycle interface, MCPServer session-delete wiring, extracted network-intercept cache module, targeted unit tests.
- Tests reviewed: direct cache isolation/eviction tests and MCPServer registration test.
- Checks considered: local jest/eslint/tsc passed; GitHub dependency-audit/lint/test passed, build pending at review time.

## Blocking Issues
None.

## Warnings
None.

## Mutation-Test Thinking
- Likely mutants that should be killed: deleting the wrong session id, ignoring empty ids, failing to register `onSessionDelete`, replacing synchronous removal with no-op.
- Mutants current tests may not catch: future composite-key device scopes beyond the current `session|...` prefix convention if the key format changes.
- Additional tests recommended: none required for this PR scope.

## Complexity / CRAP-style Risk
- High-risk functions/modules: low; cleanup is isolated in `network-intercept-cache.ts`.
- Complexity increase: minimal optional transport hook plus one cache helper.
- Test coverage concern: none for stated scope.
- Refactoring recommendation: none.

## Test Quality Assessment (6/7)
- Strong tests: prove per-session eviction preserves other session rules and MCPServer wires the transport delete hook.
- Weak tests: no live HTTP DELETE smoke, but local unit coverage is appropriate for a narrow lifecycle wiring change.
- Missing edge cases: no blocking gap.
- Mocking concerns: MCPServer transport mock is narrow and directly exercises registration behavior.

## Security / Operational Risk
None. The change reduces long-lived state retention; no auth or external input surface is widened.

## Looks Good
- Synchronous cache cleanup avoids DELETE/reuse races.
- Optional transport hook keeps stdio behavior unchanged.
- Cache extraction removes the import-cycle risk cleanly.

## Final Recommendation
Approve. Merge once the pending GitHub build check completes successfully.

shaun0927:
## Ready to merge — approve summary

All required CI checks are green on the latest commit `6a3bf234` (`lint`, `test`, `dependency-audit`, `build` — all `SUCCESS`). No P0/P1 findings from any bot (no Codex review comments on this PR after the inherited findings were addressed). Local verification on the same SHA: `npx tsc --noEmit --pretty false` exits 0, `npx eslint --max-warnings=0` on the six touched files exits 0, and `npx jest --runTestsByPath tests/unit/network-intercept-tool.test.ts tests/unit/mcp-server-session-cleanup.test.ts --runInBand` reports `Tests: 6 passed, 6 total`. This summary records why the PR is in a mergeable state.

### What this PR is

The **follow-up to #763** that was split out during the v0.6.1 release (PR #762, Codex P2 #3) because the cleanest fix needed new MCP-server-side session-lifecycle wiring. Issue #763 stayed open with full repro and a proposed fix outline; this PR implements that outline:

1. Introduces an optional `onSessionDelete(handler)` hook on the `MCPTransport` interface (`src/transports/index.ts:28-34`).
2. `MCPServer.start()` calls `this.transport.onSessionDelete?.(sessionId => removeNetworkInterceptorForSession(sessionId))` (`src/mcp-server.ts:200-202`).
3. Extracts the existing per-session interceptor cache from `src/tools/network-intercept.ts` into a new `src/tools/network-intercept-cache.ts` module so that both `mcp-server.ts` and `network-intercept.ts` can call into it without re-introducing a `mcp-server → network-intercept → mcp-server` import cycle.

The HTTP transport already exposed an `onSessionDelete` lifecycle callback (`src/transports/http.ts:65-67`); it just had no caller. After this PR it does, and the per-session `NetworkInterceptor` instance is evicted synchronously the moment `DELETE /mcp` lands.

### Scope / direction review

| Axis | Verdict |
|---|---|
| Aligned with project direction | ✅ Closes the only deferred Codex finding from #762's release; matches the existing per-session scoping pattern from develop's #753 series. |
| Single concern | ✅ One concern: evict per-session interceptor state on HTTP session deletion. Six files, +125/−21 net, no drive-by refactors beyond the cycle-breaking module extraction the new MCP-server caller forces. |
| Over-engineering risk | ✅ Considered and rejected: (a) plumbing a generic "session-lifecycle event bus" (not needed for a single subscriber — the optional hook is the simplest stable surface), (b) eviction inside `HTTPTransport` itself (would couple transport code to a tool's cache — wrong layer), (c) deferring eviction to next access via TTL (loses the determinism we want and makes test assertions hard). |
| Public API surface change | ✅ Additive only: `MCPTransport.onSessionDelete?` is optional, so stdio and any other stateless transport keep working unchanged; the new `removeNetworkInterceptorForSession(sessionId)` helper and the re-extracted `getNetworkInterceptorForSession` / `networkInterceptor` / `resetNetworkInterceptorsForTest` symbols are still re-exported from `src/tools/network-intercept.ts`, so existing imports continue to resolve. |

### Why merge is OK

- **Cycle-safe extraction.** `network-intercept-cache.ts` has zero dependencies on `mcp-server.ts` (it only imports `NetworkInterceptor`), so `mcp-server.ts → network-intercept-cache.ts` is a clean acyclic edge. `network-intercept.ts` re-exports the same symbols so call sites are unchanged.
- **Synchronous eviction matches the lifecycle.** `removeNetworkInterceptorForSession` is plain `Map.delete` work; there's no async hop between "DELETE /mcp received" and "cache entry gone", which closes the DELETE/reuse race that motivated #763.
- **Composite-key safe.** The cache key format is `<sessionId>` or `<sessionId>|<deviceId>` (the latter introduced in #762 / `fabe0ab1`). Eviction strips both `key === sessionId` and `key.startsWith(\`${sessionId}|\`)`, so the device-scoped extension stays compatible.
- **STDIO behaviour preserved.** `onSessionDelete` is `optional?:` on the `MCPTransport` interface and the call site uses `this.transport.onSessionDelete?.(...)`. The stdio transport doesn't define it, so its behaviour is byte-identical to before.
- **Defense-in-depth, not just memory hygiene.** Beyond the GC win, this prevents stale `InterceptRule`s from a previous session from being served on a freshly minted same-id reuse (an HTTP-transport corner case that would otherwise depend on identity reuse semantics).

### Verification on `6a3bf234`

```
$ npx tsc --noEmit --pretty false           # exit 0
$ npx eslint --max-warnings=0 \
    src/tools/network-intercept.ts \
    src/tools/network-intercept-cache.ts \
    src/transports/index.ts src/mcp-server.ts \
    tests/unit/network-intercept-tool.test.ts \
    tests/unit/mcp-server-session-cleanup.test.ts   # exit 0
$ npx jest --runTestsByPath \
    tests/unit/network-intercept-tool.test.ts \
    tests/unit/mcp-server-session-cleanup.test.ts --runInBand
Tests:       6 passed, 6 total
```

GitHub CI on the same commit: `lint` ✅ · `test` ✅ · `dependency-audit` ✅ · `build` ✅.

### Test coverage added

| Test | What it locks |
|---|---|
| `evicts only the selected MCP session cache on session deletion` | `removeNetworkInterceptorForSession('session-a')` returns 1, drops the `session-a` interceptor (subsequent `getNetworkInterceptorForSession('session-a')` returns a *new* instance with no rules), and leaves `session-b` rules intact. Kills the obvious "delete wrong session" / "no-op delete" mutants. |
| `reports no eviction for unknown or empty session ids` | Returns `0` for `'missing'` and `''`, doesn't touch existing entries. Kills the "delete everything on empty string" mutant. |
| `registers network interceptor eviction for transport session deletion` (MCPServer) | Confirms `MCPServer.start({ transport: 'http' })` wires `onSessionDelete(handler)` and that calling `handler('session-a')` triggers `removeNetworkInterceptorForSession('session-a')`. Kills "forgot to register" and "called with wrong sessionId" mutants. |

### Out-of-scope follow-ups (not blocking this PR)

1. **Live HTTP DELETE smoke test.** Documented gap; unit coverage on the wiring point is appropriate for a narrow lifecycle change.
2. **Other per-session caches (if any are added later).** They can register their own subscribers via the same optional hook.

Approving on the strength of green CI, zero bot findings, the scope-discipline above, and the wiring + cache-isolation tests that lock both the registration point and the eviction semantics.

shaun0927:
## PR #764: fix(network-intercept): evict session cache on HTTP delete

`/pr-review-os` skill pass — 6-axis (WebKit / iOS / Simulator / MCP / Architecture / Reliability) sweep on the latest commit `6a3bf234`.

_None._

### P1 — Must Fix (should fix in this PR)
_None._

### P2 — Improve (can be follow-up)
_None at the >= 60/100 confidence threshold._

### Per-axis findings

| Axis | Finding |
|---|---|
| **WebKit Protocol** | Not applicable — this PR is MCP-server-side session-cache plumbing. No WebKit Remote Debugging Protocol calls are added or modified. The `NetworkInterceptor` class itself is untouched; only the per-session caching layer in front of it is reorganised. |
| **iOS Safari Compatibility** | Not applicable — no DOM, no JS-injection payload changes. The interceptor's injected `fetch` / `XMLHttpRequest` override script in `src/network-interceptor.ts` is byte-identical post-PR. |
| **Simulator** | Not applicable — no `simctl` calls added. Device scoping uses the existing `device_id` parameter pattern. |
| **MCP Protocol** | ✅ `src/transports/index.ts` adds `onSessionDelete?(handler)` as an *optional* member of the `MCPTransport` interface, so stdio (and any future stateless transport) keeps working without a method-not-found at call time — the call site in `mcp-server.ts:200` uses `this.transport.onSessionDelete?.(...)`. `src/transports/http.ts:65-67` already exposed the matching method; lines 501-502 fire it on `DELETE /mcp`. No `console.log` calls in any src/ file touched by this PR (the only `console.error` use is in `http.ts:38`, which is pre-existing and correct for stderr logging). `tools/list` response shape is unchanged (the `network_intercept` tool's `inputSchema` is untouched). |
| **Architecture** | ✅ No `puppeteer-core` / `CDP` / Chrome imports. The new `src/tools/network-intercept-cache.ts` module has a single non-type import (`NetworkInterceptor`); `mcp-server.ts → network-intercept-cache.ts` is a clean acyclic edge (the new cache module never imports `mcp-server.ts`, sidestepping the `mcp-server → network-intercept → mcp-server` cycle that would have happened if cache code stayed in `network-intercept.ts`). `network-intercept.ts` re-exports the moved symbols so all existing import sites continue to resolve. Types are explicit (`number` return on `removeNetworkInterceptorForSession`, `Map<string, NetworkInterceptor>` cache); no `any`. |
| **Reliability** | ✅ `removeNetworkInterceptorForSession` is synchronous `Map.delete` work — no async hop between "DELETE /mcp received" and "cache entry gone", which closes the DELETE/reuse race that motivated #763. Composite-key safe: strips both `key === sessionId` and `key.startsWith(\`${sessionId}|\`)`, preserving the device-scoped extension introduced in #762 (`fabe0ab1`). Defensive early return on empty sessionId (`if (!sessionId) return 0;`) prevents the "delete everything on empty string" mutant. Iterating `Array.from(interceptorsBySession.keys())` instead of the live iterator avoids in-loop mutation hazards. No new resource handles are created in the eviction path; the previously cached `NetworkInterceptor` instance is dropped to GC after eviction (no `close()` / `disable()` plumbing needed — the WebKit-side rules are torn down when the session itself is gone). |

### Wiring proof

- HTTP DELETE path: `src/transports/http.ts:212 (case 'DELETE')` → `src/transports/http.ts:501-502 (this.sessionDeleteHandler(sessionId))`.
- Registration: `src/mcp-server.ts:200-202 (this.transport.onSessionDelete?.((sessionId) => removeNetworkInterceptorForSession(sessionId)))`.
- Cache eviction: `src/tools/network-intercept-cache.ts:20-31`.

End-to-end: a `DELETE /mcp` with the deleted session's id reaches `removeNetworkInterceptorForSession(sessionId)` synchronously, dropping `<sessionId>` and every `<sessionId>|<deviceId>` entry from `interceptorsBySession`. Stdio sessions skip the optional hook entirely.

### Test coverage delta (+3 unit tests, +1 wiring test)

| Test | Mutant killed |
|---|---|
| `evicts only the selected MCP session cache on session deletion` | (a) returning the wrong removed-count, (b) deleting the wrong session, (c) deleting *both* sessions on a single call, (d) returning the cached interceptor instead of a freshly constructed one for a re-acquired sessionId after delete. |
| `reports no eviction for unknown or empty session ids` | (a) deleting everything when sessionId is the empty string, (b) returning a non-zero count for unknown sessions. |
| `registers network interceptor eviction for transport session deletion` (MCPServer) | (a) forgetting to register `onSessionDelete`, (b) registering it with a closure that drops the sessionId, (c) calling `removeNetworkInterceptorForSession()` with a hard-coded constant rather than the handler argument. |
| (Existing) `clear/disable restores and clears only the selected session` | Anchors the per-session scoping the new eviction extends. |

### Verification

```
$ npx tsc --noEmit --pretty false           # exit 0
$ npx eslint --max-warnings=0 \
    src/tools/network-intercept.ts \
    src/tools/network-intercept-cache.ts \
    src/transports/index.ts src/mcp-server.ts \
    tests/unit/network-intercept-tool.test.ts \
    tests/unit/mcp-server-session-cleanup.test.ts   # exit 0
$ npx jest --runTestsByPath \
    tests/unit/network-intercept-tool.test.ts \
    tests/unit/mcp-server-session-cleanup.test.ts --runInBand
Test Suites: 2 passed, 2 total
Tests:       6 passed, 6 total
```

GitHub CI on `6a3bf234`: `lint` ✅ · `test` ✅ · `dependency-audit` ✅ · `build` ✅.

### Summary

| Priority | Count |
|----------|-------|
| P0 | 0 |
| P1 | 0 |
| P2 | 0 |

### Verdict

**APPROVE.**

Rationale, in order of weight:
1. **Zero blocking findings** on any of the six review axes; the diff is 6 files, +125/−21 net, with no drive-by changes beyond the cycle-breaking module extraction that the new MCP-server caller forces.
2. **Closes the only deferred Codex finding from the v0.6.1 release** (#762 Codex P2 #3), exactly along the fix outline that #763 captured. The wiring lands at the layer the issue body proposed: an optional transport hook, registered by `MCPServer.start()`, draining into a stateless cache helper.
3. **No layering violation introduced.** `network-intercept-cache.ts` is a leaf module (one import: `NetworkInterceptor`); `mcp-server.ts → network-intercept-cache.ts` is the new edge — clean and acyclic. The transport interface stays transport-agnostic (optional hook, stdio unaffected).
4. **Synchronous, deterministic eviction.** No timers, no setImmediate gap between session-delete and cache eviction. The race window that motivated #763 (DELETE → immediate same-id reuse → stale `NetworkInterceptor` instance served) is now zero-width.
5. **Composite-key compatible.** Strips both `<sessionId>` and `<sessionId>|<deviceId>` keys; the device-scoped extension from #762's `fabe0ab1` keeps working unchanged.
6. **Defense-in-depth on the eviction helper itself.** Early return on empty sessionId, iterator-snapshot to avoid in-loop mutation hazards, explicit `number` return type for observability.

### Merge Notes

- No file overlap with PR #765 (`fix/693-flutter-semantics-inactive-diagnostic`) — independent merge order; either can land first.
- No file overlap with PR #762 (`release/0.6.1`) for the net-new diff; the cache extraction touches files that #762 also touched (`src/tools/network-intercept.ts`) but #764 branches from `develop` after #762's `fabe0ab1` landed, so the merge resolves cleanly.
- Recommended order: **#762 (release) → #764 → #765** (or #765 → #764). All three are independently mergeable today.

— `/pr-review-os` skill, senior-engineer pass.

Issue #834, opened 2026-06-02
Attach searched-tree diagnostics to element not-found errors


shaun0927:
## Goal

The most frequent inner-loop bottleneck when driving an app is "element not found": a query fails and the tool returns a bare error, forcing the developer to **manually run `app_tree` and eyeball the difference**. On a terminal miss, automatically attach a bounded snapshot of the current tree (plus nearest label candidates) so the cause is visible in one call. This automates exactly the manual `app_tree` step a developer already performs — nothing heavier.

Diagnostic enrichment only. No change to query semantics, matching, or retry.

## SSOT / Drift Check

- **Verified — query does NOT return the tree**: `bridge.query()` (`accessibility-bridge.ts:257`) returns `AXQueryResult { matches, total, query, ambiguous, deviceContentMacOSPt }` (`ax-types.ts:66`) — **matches only**, no tree. The native bridge filters internally and does not hand the tree back. Therefore a digest **requires an explicit `bridge.dumpTree()`** (`accessibility-bridge.ts:245`) — there is no "already-fetched tree" to reuse.
- **Cost constraint**: `dumpTree` can be large/slow (`accessibility-bridge.ts:155` 10MB buffer; :184 "too many elements" timeout). The dump must be **bounded** and **failure-tolerant**.
- **SSOT for the not-found path**: `app-tap-element.ts:256-268` (the `if (!match)` structured error; already carries `query`, `labelAliases`, `index`, `timeout`). This issue **adds** a `diagnostics` field; it does not alter the existing fields.
- **SSOT for tree node shape**: `app-tree.ts` / `AXNode` (`ax-types.ts`). Reuse `AXNode` shaping for the digest — do not invent a second node format.
- **Query family sharing**: `app-query.ts`, `app-assert-element.ts`, `app-wait-for.ts` share the same miss/timeout shape; the digest helper must be shared, not re-implemented.
- **Interaction with #833**: #833 adds *activation* warnings (`warning` field); this adds *not-found* diagnostics (`diagnostics` field). Separate fields — neither overwrites the other.

## Final Implementation Scope

- [ ] Add a shared `buildNotFoundDiagnostics(deviceId, query, { maxNodes })` helper that: performs **one** bounded `dumpTree()`, truncates to `maxNodes` (default 40) `AXNode`s, and lists up to 5 nearest candidates by **case-insensitive substring** match on label/identifier (no fuzzy ranking).
[...]

Issue #795, opened 2026-05-25
SSOT: OpenSafari product direction — portable MCP-native real iOS automation, debugging, and parallel QA runtime


shaun0927:
## Thesis

OpenSafari is not just “Safari automation”. OpenSafari is a **portable MCP-native real iOS automation, debugging, and parallel QA runtime**.

It should allow Codex, Claude Code, Hermes Agent, OpenClaw, and any standards-compliant MCP client to drive real iOS Simulator workflows across Safari, WebView, Flutter iOS apps, and native UIKit / SwiftUI apps without relying on client-specific prompts, hidden assumptions, or manual simulator handling.

The product direction is not “more raw tools”. The direction is stable semantic actions, structured failures, mobile debugging evidence, parallel simulator orchestration, safe auth/session automation, and backend abstraction behind portable MCP contracts.

## Product position

OpenSafari should position itself as a portable MCP-native runtime for real iOS Simulator automation and debugging.

It is not primarily an Appium clone, a generic XCTest replacement, a production credential manager, a private-API-first framework, a Claude-only or Codex-only workflow, or a raw coordinate automation library.

Its unique value is the combination of real iOS Simulator fidelity, direct WebKit / AX / Flutter VM / simulator integration, MCP-native access for AI agents, headless CI operation, structured debugging evidence, and parallel multi-device workflows.

### 1. Portable MCP contract first

Every OpenSafari capability must be usable by any MCP client through stable tool contracts: stable tool name, explicit JSON input schema, structured JSON response, stable error code, recoverable flag where applicable, suggested next action, backend metadata, device/session identity, and evidence artifacts when relevant.

Principle: **If an MCP client cannot decide the next action from the structured response, the tool contract is incomplete.**

### 2. Semantic mobile actions over raw operations

OpenSafari should optimize for mobile user intent, not low-level input primitives. Raw coordinate tap/swipe/type remain useful, but agent-facing workflows should prefer semantic tools with verifiable postconditions such as `app_goto_screen`, `app_wait_for`, `app_tap_element`, `app_type_element`, `app_alert_handle`, `app_dismiss_overlay`, `app_pop_until`, `app_deeplink` with postcondition, `auth_restore_native`, and `workflow_*`.

Principle: **Opening a URL, sending a tap, or injecting text is not success. Reaching the expected app state is success.**

### 3. Mobile debugging is a first-class product surface

[...]

Issue #12, opened 2026-03-25
Epic 3B: QA Report Generator & Historical Tracking


shaun0927:
# Epic 3B: QA Report Generator & Historical Tracking

**Phase:** 3 — iOS QA Engine | **Dependencies:** Epic 3A | **Goal Coverage:** Goal 4

---

## Stories

| Story | Issue | Estimate | Status |
|-------|-------|----------|--------|
| 3B.1: qa_full_audit Tool & Score Calculation | #64 | 3–4h | 🔲 |
| 3B.2: Historical Tracking & Regression Detection | #66 | 2–3h | 🔲 |

---

## Objective

Create a comprehensive QA report tool that runs all detectors, aggregates results, classifies by severity, suggests fixes, and tracks regressions across runs.

## Tasks

| # | Task | Description |
|---|------|-------------|
| 3B.1 | `qa_full_audit` tool | Run all 13 detectors (parallel where possible), aggregate into single report |
| 3B.2 | Severity classification | Classify issues using defined weights (see table below) |
| 3B.3 | Fix suggestions | Actionable code snippets with before/after examples |
| 3B.4 | JSON report | Structured report: summary, per-detector results, total score |
| 3B.5 | Markdown report | Human-readable report for GitHub issues/PR comments |
| 3B.6 | **Score calculation** | Weighted score (0-100) per severity weight table |
| 3B.7 | Historical storage | Save reports to `~/.opensafari/reports/{site}/{timestamp}.json` |
| 3B.8 | **Regression detection** | Compare with previous run via issue fingerprint (detector + selector hash). Classify as: new / fixed / recurring |
| 3B.9 | CI integration | Exit code: 0 (pass), 1 (critical issues found). Configurable threshold |
| 3B.10 | **Report rotation** | Keep last 30 reports per site, auto-delete older ones |

## Severity Weight Table

| Severity | Weight | Deduction per Issue | Examples |
|----------|--------|-------------------|----------|
| Critical | 10 | -10 points | Auto-zoom on primary input, page completely broken on mobile |
| High | 5 | -5 points | Touch target too small on main CTA, horizontal overflow |
| Medium | 2 | -2 points | Missing PWA meta tag, dark mode visual difference |
| Low | 1 | -1 point | Non-critical input without inputMode, minor stacking overlap |

**Score = max(0, 100 - sum(weight × count))**

## Regression Detection Algorithm

```
For each issue in current run:
  fingerprint = hash(detector + selector + problem_type)
[...]

PR #837, opened 2026-06-02
feat(native): attach searched-tree diagnostics to not-found errors (#834 PR 4.1)

shaun0927:
## Summary

Implements **PR 4.1 of #834**. The most frequent inner-loop bottleneck is a query that fails to match, after which the developer manually re-runs `app_tree` to see what was on screen. This attaches that snapshot automatically.

On a **terminal miss** (after all relaxed/alias attempts), `app_tap_element`'s "Element not found" error now carries:
- `diagnostics`: a bounded digest of the searched tree (`searchedNodeCount`, `truncated`, up to 40 compacted `nodes`, up to 5 substring `candidates`);
- `hint`: a one-line pointer to the candidates / autoScroll.

## Design (bounded & graceful by construction)

New shared helper `src/native/not-found-diagnostics.ts`:
- **exactly one** `dumpTree`, and **only** inside the terminal-miss branch (single call site → never on success or per retry);
- hard node cap (`DEFAULT_MAX_NODES = 40`) — never an unbounded dump;
- **substring** candidate matching only (no fuzzy/Levenshtein/AI ranking);
- returns `undefined` when the dump throws/times out → original error is preserved, failure never worsened.

`AXQueryResult` carries only matches (verified — the bridge filters natively and does not return the tree), so a digest genuinely requires this one explicit `dumpTree`; there is no already-fetched tree to reuse.

## Verification

```
npx jest app-tap-element not-found-diagnostics   # 60 passed
npx tsc --noEmit                                  # exit 0
npx eslint <changed files>                        # exit 0
```

Tests assert: digest capped + truncation flag; substring candidate surfaces; empty candidates when no match (never fabricated); identifier/value matching; **dumpTree called exactly once**; **graceful `undefined` on dump failure**; and tool-level integration that diagnostics attach on a miss and are omitted when the dump fails.

## Scope guardrails

- Diagnostic enrichment only — no change to query matching, retry, or success payloads.
- No fuzzy ranking, no screenshots, no `app_tree` changes, no new required params.

Part of #834.

### DIFF
--- a/src/native/not-found-diagnostics.ts
+++ b/src/native/not-found-diagnostics.ts
@@ -1,0 +1,122 @@
+/**
+ * not-found-diagnostics — attach a bounded snapshot of the searched
+ * accessibility tree to "element not found" errors (issue #834).
+ *
+ * The most frequent inner-loop bottleneck is a query that fails to match,
+ * after which the developer manually re-runs `app_tree` to see what was
+ * actually on screen. This helper automates exactly that one step: on a
+ * terminal miss it performs a single bounded `dumpTree`, returns a capped
+ * digest plus the nearest label/identifier candidates, and degrades to
+ * `undefined` if the dump fails — so it never makes the failure worse.
+ *
+ * Deliberately minimal: substring matching only (no fuzzy/ranking), a hard
+ * node cap, and exactly one dump. See the issue's over-engineering checklist.
+ */
+
+import type { AXNode, AXQuery } from './ax-types';
+
+/** Default ceiling on how many nodes the digest carries. */
+export const DEFAULT_MAX_NODES = 40;
+
+/** Compact node shape — just enough to recognise an element, no geometry. */
+export interface CompactAXNode {
+  role: string;
+  label?: string;
+  identifier?: string;
+  value?: string;
+  path: string;
+}
+
+export interface NotFoundDiagnostics {
+  /** Total nodes in the dumped tree (may exceed `nodes.length`). */
+  searchedNodeCount: number;
+  /** Whether `nodes` was truncated at the cap. */
+  truncated: boolean;
+  /** Up to `maxNodes` compacted nodes from the searched tree. */
+  nodes: CompactAXNode[];
+  /** Up to 5 nodes whose label/identifier/value contains the query term. */
+  candidates: CompactAXNode[];
+}
+
+/** Minimal surface this helper needs from the accessibility bridge. */
+export interface TreeDumper {
+  dumpTree(options?: { deviceId?: string; maxDepth?: number }): Promise<AXNode>;
+}
+
+const MAX_CANDIDATES = 5;
+
+function compact(node: AXNode): CompactAXNode {
+  return {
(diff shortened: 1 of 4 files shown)


shaun0927:
## Analysis & merge-readiness review (automated)

**Intent vs OpenSafari direction (#795):** Implements #834 PR 4.1 and directly serves Pillar 3 *"mobile debugging is a first-class product surface — every failed high-level action should be explainable without manually rerunning the simulator flow."* Adds `buildNotFoundDiagnostics()` and wires it into `app_tap_element`'s terminal not-found error so the developer sees the searched tree + nearest candidates in one call instead of a manual `app_tree` round-trip. ✅

**Deep analysis — verified correct:**
- **Single bounded dump:** exactly one `dumpTree` inside try/catch; returns `undefined` on throw so it never worsens the failing path.
- **Traversal:** iterative DFS (`stack.pop()`) visits each node once over a finite AX tree (no cycles → terminates); `nodes` hard-capped at 40, `candidates` at 5, `searchedNodeCount`/`truncated` accurate. Empty query term → no candidates (no fabrication).
- **Bounded payload:** `compact()` strips geometry/children/traits — worst case ~45 short-string records. No unbounded memory.
- **Tests:** truncation, multi-field (label/identifier/value) candidate matching, empty-candidates, cap-at-5, and graceful-undefined-on-dump-failure all covered.

**Code review verdict:** MERGE-READY — 0 P0/P1. CI green.

Merging as the foundation for #840 (PR 4.2).

shaun0927:
## Post-merge review analysis — ⚠️ a P0 redaction gap shipped; fix in #841

This PR was **merged at 08:14 before a code-review fix could land on the branch**, so `develop` currently ships the not-found `diagnostics` digest **without redaction**. The corrective is **#841** (cherry-picks the fix onto current `develop`).

**Intent & #795 alignment (sound).** The `diagnostics` field on "element not found" directly serves Pillar 3 — *"Every failed high-level action should be explainable without manually rerunning the simulator flow."* One bounded dump, capped digest, nearest substring candidates, graceful on failure. Correctness verified: exactly one `dumpTree`, only on terminal miss, DFS is acyclic/terminating.

**P0 found (now fixed in #841), #795 #12 / pillar 6 — secrets redaction:**
The digest emitted each node's `value` (the `AXValue`, i.e. **user-entered text** — passwords, emails, tokens, OTPs) directly into the MCP response body. The existing `debug_bundle_collect` deliberately avoids this (AX tree goes to a local file; logs/env run through `redactText()`). #841:
- never emits `value` (kept only for internal matching);
- redacts `label`/`identifier` through the shared credential redactor;
- adds `redactionPolicy` to signal posture;
- also bounds the dump (`maxDepth: 8`) and switches DFS to pre-order.

**P1 (acceptable):** keeping `diagnostics` as a lightweight always-on **redacted** complement to the heavy opt-in `debugBundle` is fine — different cost profiles, redaction parity achieved.

Net: the feature is well-aligned with #795; merge-readiness is restored by **#841**.

Issue #833, opened 2026-06-02
Surface silent degradations in app_tap_element as actionable warnings


shaun0927:
## Goal

Semantics-dependent native tools degrade silently when the accessibility tree fails to populate: five tools call `ensureSemanticsActive(...)` and **ignore its boolean result**, then query an empty tree and return a bare "not found" with no cause. Close this across the family with one shared warning helper, and separately surface the silent raw-coordinate fallback in `app_tap_element`.

Warnings only. No change to dispatch, query, or coordinate math; no new tool parameters.

## SSOT / Drift Check

- **Existing-good pattern to mirror (do NOT modify)**: `app-tap.ts:143,282-289` already captures the activation result and emits a structured `warning`; `app-tree.ts:17-22,100` already raises structured `DEVICE_CONTENT_ROOT_EMPTY` with retry. These are the reference, not targets.
- **Silent call sites to fix (verified ignore the result)**: `app-tap-element.ts:181`, `app-query.ts:93`, `app-wait-for.ts:189`, `app-type-element.ts:223`, `app-assert-element.ts:123`.
- **Warning channel SSOT**: each tool already returns a `warning` field — reuse each tool's existing channel; no new channel.
- **Coordinate fallback SSOT**: `app-tap-element.ts:637` `getIosPtSizeForDevice` returns `null` for **both** preset-miss (permanent, cached) and transient simctl failure (un-cached early return, :645-651). The accuracy warning must fire **only on preset-miss**.
- **Note on false positives**: `ensureSemanticsActive` already polls/retries to a deadline (`semantics-activator.ts`), so a `false` return means the retry window was exhausted — the warning is not premature. Wording must still be non-absolute ("may not have populated"), since a slow device can recover on a later call.
- **Drift risk**: copy-pasting the message into five files = five drifting contracts ⇒ one shared helper is the single source.

## Final Implementation Scope

- [ ] **C1 — shared helper**: add `activateSemanticsOrWarn(deviceId, opts)` returning `{ active: boolean, warning?: string }`; the `warning` states the cause + remediation (rebuild `--debug` / restart app / add `Semantics()` wrapper) in non-absolute wording.
- [ ] **C1 — wire 5 call sites**: replace the result-ignoring `await ensureSemanticsActive(...)` at the five verified sites with the helper; merge any `warning` into that tool's existing `warning` field.
- [ ] **C1 — explicitly leave `app_tap` and `app_tree` untouched** (already handled; no refactor onto the helper in this issue).
[...]

PR #839, opened 2026-06-02
feat(app-tap-element): warn on raw-coordinate fallback for unknown device presets (#833 PR 3.2)

shaun0927:
## Summary

Implements **PR 3.2 of #833**. When `app_tap_element` resolves a device whose name is not in `DEVICE_PRESETS`, no macOS-pt→iOS-pt scale factor is known, so it taps with **raw AX-frame coordinates** — which are off (≈1.7×) on a scaled Simulator window. This was silent; it now warns.

## Change

`getIosPtSizeForDevice` now returns a discriminated `{ size, reason }`:
- `preset_miss` — device resolved but not in the preset table → emit a coordinate-accuracy `warning`;
- `transient` — simctl could not see the device → **stay silent** (may recover on a later call; avoids alarm fatigue) and remains **un-cached**;
- `ok` — usable size → convert as before.

Warning-only: coordinate math and dispatch are unchanged; raw coordinates are forwarded exactly as before.

## Verification

```
npx jest app-tap-element   # 52 passed
npx tsc --noEmit           # exit 0
npx eslint <changed files> # exit 0
```

Tests assert: preset-miss device → `warning` contains "not in the known preset table"; transient (`getDevice` returns null) → **no** preset-miss warning (the alarm-fatigue guard); `ok` preset → conversion still applied (existing tests unchanged).

## Scope guardrails

- Warning-only — no scale-factor guessing for unknown presets, no dispatch change.
- Preserves `getIosPtSizeForDevice` caching (transient stays un-cached).

Part of #833.

### DIFF
--- a/src/tools/app-tap-element.ts
+++ b/src/tools/app-tap-element.ts
@@ -490,9 +490,10 @@ export function registerAppTapElementTool(server: MCPServer): void {
         // — same behavior as before. The lookup is deferred to here (rather
         // than running before AXPress) so successful AXPress taps avoid the
         // `simctl list` round-trip entirely.
+        let coordinateWarning: string | undefined;
         const macOSPtSize = queryResult?.deviceContentMacOSPt;
         if (macOSPtSize) {
-          const iosPtSize = await getIosPtSizeForDevice(deviceId);
+          const { size: iosPtSize, reason } = await getIosPtSizeForDevice(deviceId);
           if (iosPtSize) {
             const before = { x: centerX, y: centerY };
             const converted = convertMacOSPtToIOSPt(before, macOSPtSize, iosPtSize);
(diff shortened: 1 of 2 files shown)


shaun0927:
## Analysis & merge-readiness review (automated)

**Intent vs OpenSafari direction (#795):** Directly serves the P1 *element-interaction ladder* requirement — *"verified coordinate fallback only when proof is available."* When no preset scale factor exists (`preset_miss`), the tap can no longer convert macOS-pt → iOS-pt, so it now warns instead of silently dispatching raw (≈1.7× off) coordinates. Consistent with #833 (surface silent degradations as actionable warnings). ✅

**Deep analysis — verified correct:**
- **Caching:** the migration from `Map<string, Size2D|null>` (`.has()` + `?? null`) to `Map<string, IosPtSizeResult>` (truthiness check) is regression-free. `transient` is never `.set()` (both `!device` and `catch` return early un-cached → recovers on a later call); `ok`/`preset_miss` are cached objects (truthy) → returned immediately. Permanent preset-miss / recoverable-transient semantics preserved.
- **Warning gating:** `coordinateWarning` fires **only** on `reason === 'preset_miss'`, is appended to the existing `warnings[]`, and joined into `response.warning`. `transient` and `ok` never raise it (alarm-fatigue guard intact).
- **Types:** discriminated `{ size, reason }` is type-safe at the single call site (null-check then reason-check, no narrowing dependency). `tsc --noEmit` clean.
- **Tests:** all three branches covered — `ok` converts (no warning), `preset_miss` warns ("not in the known preset table"), `transient` does **not** warn.

**Code review verdict:** MERGE-READY — 0 P0/P1/P2. CI green (build/lint/test/dependency-audit).

Merging.

